PR #3, opened 2021-11-09
docs: load the new examples

vladfrangu:


website/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "scripts": {
         "examples": "docusaurus-examples",
-        "start": "rimraf .docusaurus && docusaurus start",
+        "start": "rimraf .docusaurus && npm run generate && docusaurus start",
         "build": "rimraf .docusaurus && docusaurus build",
```
B4nan, line 5: i guess we should move the `generate` command here as well (and maybe to `deploy` script too, and remove it from the `docs` workflow)

B4nan (approved):
otherwise lgtm

website/package.json
vladfrangu: Agreed, have done that

B4nan: now we can also remove the `npm run generate` call from `docs.yml`

vladfrangu: Forgot to do that, woops. Committed the removal

PR #4, opened 2021-11-10
feat: add generics to Datasets and KVStore

vladfrangu:


packages/apify/src/storages/dataset.ts
```diff
@@ -227,9 +225,10 @@ export class Dataset {
      *   If `true` then the function doesn't return empty items.
      *   Note that in this case the returned number of items might be lower than limit parameter and pagination must be done using the `limit` value.
      */
-    async getData<T = unknown>(options = {}): Promise<DatasetContent<T>> {
+    async getData(options = {}): Promise<DatasetContent<Data>> {
```
B4nan, line 228: let's add a TODO here so we know we need to create the options interface, and for now at least use `Dictionary`

packages/apify/src/storages/dataset.ts
```diff
@@ -227,9 +225,10 @@ export class Dataset {
...
         try {
-            // @ts-ignore
+            // TODO: in apify_client, DatasetClient#listItems needs to take in a generic for the return type
```
B4nan, line 230: you could look into that as the next step

packages/apify/src/storages/dataset.ts
```diff
@@ -262,8 +261,7 @@ export class Dataset {
      * }
      * ```
      */
-    async getInfo<T extends Dictionary = Dictionary>(): Promise<T> {
-        // @ts-ignore
+    async getInfo() {
```
B4nan, line 264: would be good to specify the type explicitly as this is public api

packages/apify/src/storages/key_value_store.ts
```diff
@@ -287,7 +287,7 @@ export class KeyValueStore {
      * @param {string} [options.exclusiveStartKey] All keys up to this one (including) are skipped from the result.
      * @return {Promise<void>}
      */
-    async forEachKey(iteratee, options = {}) {
+    async forEachKey(iteratee: KeyConsumer, options = {}) {
```
B4nan, line 290:
```suggestion
    async forEachKey(iteratee: KeyConsumer, options: Dictionary = {}) {
```

packages/apify/src/storages/key_value_store.ts
```diff
@@ -298,7 +298,7 @@ export class KeyValueStore {
      * @return {Promise<Promise<void> | undefined>}
      * @private
      */
-    async _forEachKey(iteratee, options = {}, index = 0) {
+    private async _forEachKey(iteratee: KeyConsumer, options = {}, index = 0): Promise<void> {
```
B4nan, line 301:
```suggestion
    private async _forEachKey(iteratee: KeyConsumer, options: Dictionary = {}, index = 0): Promise<void> {
```

B4nan (changes requested):
looking good, just few nitpicks

packages/apify/src/storages/dataset.ts
vladfrangu: It felt redundant since TS infers the type anyways,but I can add it back! Can we setup the eslint rule then that always enforces return types?

B4nan: for public methods yes. the thing is that with inference, the upstream can change return value and introduce a BC. (it probably won't happen with this one, more like a general rule of thumb)

packages/apify/src/storages/dataset.ts
```diff
@@ -132,11 +133,15 @@ export function chunkBySize(items: string[], limitBytes: number): string[] {
  * ```
  * @category Result Stores
  */
-export class Dataset {
+export class Dataset<Data extends Dictionary = Dictionary> {
     id: string;
+
```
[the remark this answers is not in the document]
vladfrangu: That's..so weird.. Before rebasing it wasn't even erroring, now it is. I'll update my deps and revert the new lines

B4nan:
@AndreyBykov take a look at this as well as #5, I guess you don't know much about generics so those could be nice learning materials for you. Feel free to comment on the code and ask things.

B4nan (approved):

PR #5, opened 2021-11-11
feat: generic pages in contexts

vladfrangu:
TODO: responses too

### DIFF
--- a/packages/apify/src/crawlers/basic_crawler.ts
+++ b/packages/apify/src/crawlers/basic_crawler.ts
@@ -16,7 +16,7 @@ import { Request } from '../request';
 import { RequestList } from '../request_list';
 import { RequestQueue } from '../storages/request_queue';
 import { Session } from '../session_pool/session';
-import { Hook } from './browser_crawler';
+import { BrowserHook } from './browser_crawler';
 import { Awaitable } from '../typedefs';
 
 export interface CrawlingContext extends Record<PropertyKey, unknown> {
@@ -678,7 +678,8 @@ export class BasicCrawler {
         }
     }
 
-    protected async _executeHooks(hooks: Hook[], ...args: Parameters<Hook>) {
+    // eslint-disable-next-line space-before-function-paren -- looks like a rule colliding with TS
+    protected async _executeHooks<Hooks extends (...args: unknown[]) => Awaitable<void>>(hooks: Hooks[], ...args: Parameters<Hooks>) {
         if (Array.isArray(hooks) && hooks.length) {
             for (const hook of hooks) {
                 await hook(...args);
--- a/packages/apify/src/crawlers/basic_crawler.ts
+++ b/packages/apify/src/crawlers/basic_crawler.ts
@@ -19,24 +19,25 @@ import { Session } from '../session_pool/session';
 import { BrowserHook } from './browser_crawler';
 import { Awaitable } from '../typedefs';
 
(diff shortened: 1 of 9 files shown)


packages/apify/src/crawlers/basic_crawler.ts
```diff
@@ -678,7 +678,8 @@ export class BasicCrawler {
         }
     }
 
-    protected async _executeHooks(hooks: Hook[], ...args: Parameters<Hook>) {
+    // eslint-disable-next-line space-before-function-paren -- looks like a rule colliding with TS
```
B4nan, line 681: yop https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/space-before-function-paren.md

B4nan: should be fixed in master

packages/apify/src/crawlers/basic_crawler.ts
```diff
@@ -678,7 +678,8 @@ export class BasicCrawler {
...
+    protected async _executeHooks<Hooks extends (...args: unknown[]) => Awaitable<void>>(hooks: Hooks[], ...args: Parameters<Hooks>) {
```
B4nan, line 682: the type arg is called `Hooks` but it describes single hook, right?

packages/apify/src/crawlers/browser_crawler.ts
```diff
@@ -17,17 +17,20 @@ import { BrowserLaunchContext } from '../browser_launchers/browser_launcher';
...
     GoToOptions extends Record<string, any> = Record<string, any>
-> = (crawlingContext: T, gotoOptions: GoToOptions) => Awaitable<void>;
-export type BrowserHook = Hook<BrowserCrawlingContext>;
-export type BrowserHandlePageFunction = (context: BrowserCrawlingContext) => Awaitable<void>;
-export type GotoFunction = (context: BrowserCrawlingContext, gotoOptions: Record<string, any>) => Awaitable<any>;
+> = (crawlingContext: Context, gotoOptions: GoToOptions) => Awaitable<void>;
+export type BrowserHandlePageFunction<Context = BrowserCrawlingContext> = (context: Context) => Awaitable<void>;
+export type GotoFunction<
+    Context = BrowserCrawlingContext,
+    GoToOptions extends Record<string, any> = Record<string, any>
```
B4nan, line 32: can we use `Dictionary` here or do we need the keys to be `any`?

B4nan (approved):
good job, just few nitpicks

packages/apify/src/crawlers/basic_crawler.ts
vladfrangu: Yeah, I'll rename this type arg to `HookLike`

packages/apify/src/crawlers/cheerio_crawler.ts
```diff
@@ -335,6 +335,7 @@ export interface CheerioHandlePageInputs extends CrawlingContext {
      */
     body: (string | Buffer);
 
+    // TODO: maybe consider making this `unknown` instead to force casting for TS users
```
B4nan, line 338: why not make it another generic argument (and default to unknown)

B4nan (approved):

packages/apify/src/crawlers/cheerio_crawler.ts
vladfrangu: Can do that, sure

PR #6, opened 2021-11-15
chore: Add ts type definitions (instead of jsdoc)

AndreyBykov:
First portion of TS type definitions, jsdoc cleaned up


packages/apify/src/autoscaling/snapshotter.ts
```diff
@@ -343,12 +342,9 @@ export class Snapshotter {
 
     /**
      * Creates a snapshot of current event loop delay.
-     * @param {Function} intervalCallback
-     * @ignore
-     * @protected
      * @internal
      */
-    _snapshotEventLoop(intervalCallback) {
+    _snapshotEventLoop(intervalCallback: () => unknown) { // TODO: check whether it should be public or protected
```
vladfrangu, line 347: It should be protected or even private

packages/apify/src/autoscaling/snapshotter.ts
```diff
@@ -444,13 +439,9 @@ export class Snapshotter {
...
      * number of requests and do not necessarily signify API
      * overloading.
-     *
-     * @param intervalCallback
-     * @ignore
-     * @protected
      * @internal
      */
-    _snapshotClient(intervalCallback) {
+    _snapshotClient(intervalCallback: () => unknown) { // TODO: check whether it should be public or protected
```
vladfrangu, line 444: Ditto

packages/apify/src/crawlers/cheerio_crawler.ts
```diff
@@ -926,3 +919,10 @@ export class CheerioCrawler extends BasicCrawler {
         }
     }
 }
+
+interface RequestFunctionOptions {
```
vladfrangu, line 923: Should probably export this? CC @B4nan thoughts?

packages/apify/src/crawlers/statistics.ts
```diff
@@ -128,11 +128,9 @@ export class Statistics {
 
     /**
      * Mark job as finished and sets the state
-     *
-     * @param {number|string} id
      * @ignore
      */
-    finishJob(id) {
+    finishJob(id: number|string) {
```
vladfrangu, line 133:
```suggestion
    finishJob(id: number | string) {
```

packages/apify/src/crawlers/statistics.ts
```diff
@@ -146,11 +144,9 @@ export class Statistics {
 
     /**
      * Mark job as failed and sets the state
-     *
-     * @param {number|string} id
      * @ignore
      */
-    failJob(id) {
+    failJob(id: number|string) {
```
vladfrangu, line 149:
```suggestion
    failJob(id: number | string) {
```

packages/apify/src/enqueue_links/click_elements.ts
```diff
@@ -15,6 +15,85 @@ import { Dictionary } from '../typedefs';
...
+     *   or an array of strings or RegExps or plain Objects from which the {@link PseudoUrl}s can be constructed.
+     *
+     *   The plain objects must include at least the `purl` property, which holds the pseudo-URL string or RegExp.
+     *   All remaining keys will be used as the `requestTemplate` argument of the {@link PseudoUrl} constructor,
+     *   which lets you specify special properties for the enqueued {@link Request} objects.
+     *
+     *   If `pseudoUrls` is an empty array, `null` or `undefined`, then the function
+     *   enqueues all links found on the page.
+     */
+    pseudoUrls?: string|RegExp|Record<string, unknown>[],
```
vladfrangu, line 46: I think we have a type for this, PseudoUrlInput?

packages/apify/src/enqueue_links/click_elements.ts
```diff
@@ -201,13 +222,8 @@ export async function clickElementsAndInterceptNavigationRequests(options): Prom
...
 
-/**
- * @param {Page} page
- * @param {Set<*>} requests
- * @return {Function}
- * @ignore
- */
-function createInterceptRequestHandler(page, requests) {
+function createInterceptRequestHandler(page: Page, requests: Set<string>): (req: PuppeteerRequest) => Promise<unknown> | undefined {
+    // @ts-ignore // TODO: Not all code paths return a value.
```
vladfrangu, line 226: Please don't use a ts-ignore directive for that
Your return type shouldn't need the `| undefined` right?

packages/apify/src/enqueue_links/click_elements.ts
```diff
@@ -226,24 +242,12 @@ function createInterceptRequestHandler(page, requests) {
...
 }
 
-/**
- * @param {Page} page
- * @param {Set<*>} requests
- * @return {Function}
- * @ignore
- */
-function createTargetCreatedHandler(page, requests) {
+function createTargetCreatedHandler(page: Page, requests: Set<string>): (target: Target) => Promise<unknown>| undefined {
```
vladfrangu, line 250:
```suggestion
function createTargetCreatedHandler(page: Page, requests: S1et<string>): (target: Target) => Promise<void> {
```

packages/apify/src/enqueue_links/click_elements.ts
```diff
@@ -201,13 +222,8 @@ export async function clickElementsAndInterceptNavigationRequests(options): Prom
...
 }
 
-/**
- * @param {Page} page
- * @param {Set<*>} requests
- * @return {Function}
- * @ignore
- */
-function createInterceptRequestHandler(page, requests) {
+function createInterceptRequestHandler(page: Page, requests: Set<string>): (req: PuppeteerRequest) => Promise<unknown> | undefined {
```
vladfrangu, line 225:
```suggestion
function createInterceptRequestHandler(page: Page, requests: Set<string>): (req: PuppeteerRequest) => Promise<unknown> {
```

vladfrangu (changes requested):
Looks great, just a few small nits

packages/apify/src/crawlers/basic_crawler.ts
```diff
@@ -39,6 +39,13 @@ export interface CrawlingContext extends Record<PropertyKey, unknown> {
     response?: IncomingMessage;
 }
 
+export interface HandleFailedRequestInput extends CrawlingContext {
+    /**
+     *   The Error thrown by `handleRequestFunction`.
```
B4nan, line 44:
```suggestion
     * The Error thrown by `handleRequestFunction`.
```

packages/apify/src/crawlers/cheerio_crawler.ts
B4nan: idk, given its used only in protected method probably not?

btw please use `;` instead of `,` inside types. i just added eslint rule to master, it has a fixer, so you can just rebase the changes from master and run `lint --fix`

packages/apify/src/enqueue_links/click_elements.ts
```diff
@@ -15,6 +15,85 @@ import { Dictionary } from '../typedefs';
 
 const STARTING_Z_INDEX = 2147400000;
 
+export interface EnqueueLinksByClickingElementsOptions {
+    /**
+     *   Puppeteer [`Page`](https://pptr.dev/#?product=Puppeteer&show=api-class-page) object.
```
B4nan, line 20: lets remove the redundant spacing (general note for the whole PR)

```suggestion
     * Puppeteer [`Page`](https://pptr.dev/#?product=Puppeteer&show=api-class-page) object.
```

packages/apify/src/crawlers/cheerio_crawler.ts
AndreyBykov: kept without export. linting - done.

packages/apify/src/enqueue_links/click_elements.ts
AndreyBykov: Hopefully did not miss it anywhere, removed in files in current PR

AndreyBykov:
Ok, I think all the mentioned changes are fixed. Regarding @ts-ignore vs @ts-expect-error - kept @ts-ignore for `Object is possibly 'null'`, as using expect-error fails the test with `Unused '@ts-expect-error' directive.`. Otherwise should be good to go, tests pass.

packages/apify/src/autoscaling/snapshotter.ts
```diff
@@ -374,13 +370,12 @@ export class Snapshotter {
...
      * Creates a snapshot of current CPU usage
      * using the Apify platform `systemInfo` event.
-     * @param {Object} systemInfo
-     * @ignore
-     * @protected
      * @internal
      */
-    _snapshotCpuOnPlatform(systemInfo) {
+    protected _snapshotCpuOnPlatform(systemInfo: SystemInfo) {
+        // @ts-expect-error // TODO Figure out 'property does not exist'
```
B4nan, line 376: define them in the interface, we can mark them as `@internal` to not have them in the docs. or maybe better document them in the interface via jsdoc saying those are platform only properties, idk, `@internal` could be enough.

packages/apify/src/enqueue_links/click_elements.ts
```diff
@@ -159,15 +183,13 @@ export async function enqueueLinksByClickingElements(options) {
  * Clicks all elements of given page matching given selector.
  * Catches and intercepts all initiated navigation requests and opened pages.
  * Returns a list of all target URLs.
- *
- * @param {object} options
- * @param {Page} options.page
- * @param {string} options.selector
- * @param {number} [options.waitForPageIdleMillis]
- * @param {number} [options.maxWaitForPageIdleMillis]
- * @ignore
```
B4nan, line 168: keep the `@ignore` comment, that is for typedoc so it is not part of the API documentation

packages/apify/src/enqueue_links/click_elements.ts
```diff
@@ -201,14 +223,8 @@ export async function clickElementsAndInterceptNavigationRequests(options): Prom
     return serializedRequests.map((r) => JSON.parse(r));
 }
 
-/**
- * @param {Page} page
- * @param {Set<*>} requests
- * @return {Function}
- * @ignore
```
B4nan, line 208: keep the @ignore comment, that is for typedoc so it is not part of the API documentation

packages/apify/src/enqueue_links/click_elements.ts
```diff
@@ -226,24 +242,12 @@ function createInterceptRequestHandler(page, requests) {
     };
 }
 
-/**
- * @param {Page} page
- * @param {PuppeteerRequest} req
- * @return {boolean}
- * @ignore
```
B4nan, line 233: keep the @ignore comment, that is for typedoc so it is not part of the API documentation

packages/apify/src/enqueue_links/click_elements.ts
```diff
@@ -226,24 +242,12 @@ function createInterceptRequestHandler(page, requests) {
...
+function isTopFrameNavigationRequest(page: Page, req: PuppeteerRequest): boolean {
     return req.isNavigationRequest()
         && req.frame() === page.mainFrame();
 }
 
-/**
- * @param {Page} page
- * @param {Set<*>} requests
- * @return {Function}
- * @ignore
```
B4nan, line 244: keep the @ignore comment, that is for typedoc so it is not part of the API documentation

packages/apify/src/enqueue_links/click_elements.ts
```diff
@@ -253,6 +257,7 @@ function createTargetCreatedHandler(page, requests) {
         // possible errors like target closed.
         try {
             const createdPage = await target.page();
+            // @ts-ignore TODO: Object is possibly 'null'. Using @ts-expect-error fails the test
```
vladfrangu, line 260: Btw, if we know the createdPage prop is always present, you can add a `!` in front of the `.` to tell TS that it cannot be null

vladfrangu (approved):

packages/apify/src/enqueue_links/click_elements.ts
```diff
@@ -253,6 +257,7 @@ function createTargetCreatedHandler(page, requests) {
...
             await createdPage.close();
```
B4nan, line 261: You can use [non-null-assertion-operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator) if you know something will be there but the type says its optional, no need for ts-ignore.

```suggestion
            await createdPage!.close();
```

packages/apify/src/enqueue_links/click_elements.ts
```diff
@@ -382,14 +381,13 @@ function updateElementCssToEnableMouseClick(el, zIndex) {
...
- * @param {Page} options.page
- * @param {number} options.waitForPageIdleMillis
- * @param {number} options.maxWaitForPageIdleMillis
  * @ignore
  */
-async function waitForPageIdle({ page, waitForPageIdleMillis, maxWaitForPageIdleMillis }): Promise<void> {
+async function waitForPageIdle({ page, waitForPageIdleMillis, maxWaitForPageIdleMillis }: {
+    page: Page;
+    waitForPageIdleMillis?: number;
+    maxWaitForPageIdleMillis?: number;
```
B4nan, line 389: let's extract this to an interface

packages/apify/src/enqueue_links/click_elements.ts
```diff
@@ -423,6 +421,7 @@ async function waitForPageIdle({ page, waitForPageIdleMillis, maxWaitForPageIdle
         timeout = activityHandler(); // We call this once manually in case there would be no requests at all.
         page.on('request', activityHandler);
         page.on('framenavigated', activityHandler);
+        // @ts-expect-error TODO: Argument of type '"targetcreated"' is not assignable to parameter of type 'keyof PageEventObject'.
         page.on('targetcreated', newTabTracker);
```
B4nan, line 425: you could type cast, not great but if its some nonstandard event, better to do that instead ts-expect-error 

```suggestion
        page.on('targetcreated' as keyof PageEventObject, newTabTracker);
```

packages/apify/src/enqueue_links/click_elements.ts
AndreyBykov: I am not sure, but I think it could actually be `null`, looking at the comment above:
// We want to close the page but don't care about
// possible errors like target closed.
I gues for some internal browser/network issues, and that's exactly the reason why it's wrapper into try/catch, but I might be wrong.

B4nan (changes requested):
image

packages/apify/src/browser_launchers/browser_launcher.ts
```diff
@@ -151,28 +151,15 @@ export abstract class BrowserLauncher<
...
         return process.env[ENV_VARS.CHROME_EXECUTABLE_PATH] || getTypicalChromeExecutablePath();
     }
 
-    /**
-     *
-     * @param {string} proxyUrl
-     * @private
-     */
-    _validateProxyUrlProtocol(proxyUrl) {
+    protected _validateProxyUrlProtocol(proxyUrl: string | undefined): void {
```
B4nan, line 162:
```suggestion
    protected _validateProxyUrlProtocol(proxyUrl?: string): void {
```

packages/apify/src/autoscaling/snapshotter.ts
```diff
@@ -374,28 +370,25 @@ export class Snapshotter {
...
-     * @param {Object} systemInfo
-     * @ignore
-     * @protected
      * @internal
      */
-    _snapshotCpuOnPlatform(systemInfo) {
+    protected _snapshotCpuOnPlatform(systemInfo: SystemInfo) {
         const { cpuCurrentUsage, isCpuOverloaded } = systemInfo;
-        const createdAt = (new Date(systemInfo.createdAt));
+        const createdAt = (new Date(systemInfo.createdAt!));
```
B4nan, line 377:
```suggestion
        const createdAt = new Date(systemInfo.createdAt!);
```

packages/apify/src/enqueue_links/click_elements.ts
```diff
@@ -355,18 +362,24 @@ export async function clickElements(page, selector) {
 /* istanbul ignore next */
 /**
  * This is an in browser function!
- * @param {Element} el
- * @param {number} zIndex
  */
-function updateElementCssToEnableMouseClick(el, zIndex) {
+function updateElementCssToEnableMouseClick(el: Element, zIndex: number): void {
```
B4nan, line 366: type this as `HTMLElement`, that way there is no need for the ignore comments. also `el.style.top/left` should be numbers (you will see that when you use `HTMLElement`)

```suggestion
function updateElementCssToEnableMouseClick(el: HTMLElement, zIndex: number): void {
```

B4nan (changes requested):
final round of nitpicks, ready to merge once those are addressed

packages/apify/src/enqueue_links/click_elements.ts
AndreyBykov: @B4nan did not get this one - TS expects zIndex/left/top to be strings. Found somewhere at SO that they are implicitly converted to strings. Should the error be expected in this case or I should just convert them to string?

B4nan (approved):

PR #10, opened 2021-11-26
chore: add more ts type definitions

AndreyBykov:
Second portion of TS type definitions, jsdoc cleaned up + updated the tests where necessary.

Main question I have here is related to the dataset. There's `DatasetDataOptions` interface, which has all params like `limit`, `skipHidden`, `fields` and so on. And there are methods such as `forEach` and some others. `forEach` has 3 options exposed (desc, fields, unwind). While inside of the method there are checks for `format`, `limit`, maybe some others. So I could basically use the `DatasetDataOptions` interface there, but in order to hide some params, I am using `DatasetIteratorOptions extends DatasetDataOptions` with some params overridden with `@internal`. Is it the correct way to do it or what's the proper way in such cases?

P.S.: sry for the delay, will start with the remaining part right away.

### DIFF
--- a/packages/apify/src/session_pool/session.ts
+++ b/packages/apify/src/session_pool/session.ts
@@ -271,11 +271,11 @@ export class Session {
      * that the target website is blocking us. This function helps to do this conveniently
      * by retiring the session when such code is received. Optionally the default status
      * codes can be extended in the second parameter.
-     * @param statusCode {number} - HTTP status code
-     * @param [blockedStatusCodes] {number[]} - Custom HTTP status codes that means blocking on particular website.
+     * @param statusCode HTTP status code.
+     * @param [blockedStatusCodes] Custom HTTP status codes that means blocking on particular website.
      * @return {boolean} whether the session was retired.
      */
-    retireOnBlockedStatusCodes(statusCode, blockedStatusCodes = []) {
+    retireOnBlockedStatusCodes(statusCode: number, blockedStatusCodes: number[] = []): boolean {
         const isBlocked = STATUS_CODES_BLOCKED.concat(blockedStatusCodes).includes(statusCode);
         if (isBlocked) {
             this.retire();
@@ -340,7 +340,7 @@ export class Session {
 
     /**
      * Transforms puppeteer cookie to tough-cookie.
-     * @param puppeteerCookie Cookie from puppeteer `page.cookies method.
+     * @param puppeteerCookie Cookie from puppeteer `page.cookies` method.
      * @internal
      */
     protected _puppeteerCookieToTough(puppeteerCookie: PuppeteerCookie): Cookie {
--- a/packages/apify/src/session_pool/session.ts
+++ b/packages/apify/src/session_pool/session.ts
@@ -170,11 +170,10 @@ export class Session {
     }
 
     /**
-     * indicates whether the session is blocked.
+     * Indicates whether the session is blocked.
      * Session is blocked once it reaches the `maxErrorScore`.
-     * @return {boolean}
      */
-    isBlocked() {
+    isBlocked(): boolean {
         return this.errorScore >= this.maxErrorScore;
     }
 
@@ -182,27 +181,24 @@ export class Session {
      * Indicates whether the session is expired.
      * Session expiration is determined by the `maxAgeSecs`.
      * Once the session is older than `createdAt + maxAgeSecs` the session is considered expired.
-     * @return {boolean}
      */
-    isExpired() {
+    isExpired(): boolean {
         return this.expiresAt <= new Date();
     }
 
     /**
      * Indicates whether the session is used maximum number of times.
      * Session maximum usage count can be changed by `maxUsageCount` parameter.
-     * @return {boolean}
      */
-    isMaxUsageCountReached() {
+    isMaxUsageCountReached(): boolean {
         return this.usageCount >= this.maxUsageCount;
     }
 
     /**
      * Indicates whether the session can be used for next requests.
      * Session is usable when it is not expired, not blocked and the maximum usage count has not be reached.
-     * @return {boolean}
      */
-    isUsable() {
+    isUsable(): boolean {
         return !this.isBlocked() && !this.isExpired() && !this.isMaxUsageCountReached();
     }
 
@@ -222,11 +218,13 @@ export class Session {
 
     /**
      * Gets session state for persistence in KeyValueStore.
-     * @return {SessionState} represents session internal state.
+     * @returns Represents session internal state.
      */
-    getState() {
+    getState(): SessionState {
         return {
             id: this.id,
+            // @ts-expect-error Type 'Serialized' is missing the following properties from type 'CookieJar':
+            // setCookie, setCookieSync, getCookies, getCookiesSync, and 11 more.
             cookieJar: this.cookieJar.toJSON(),
             userData: this.userData,
             maxErrorScore: this.maxErrorScore,
@@ -273,7 +271,7 @@ export class Session {
      * codes can be extended in the second parameter.
      * @param statusCode HTTP status code.
      * @param [blockedStatusCodes] Custom HTTP status codes that means blocking on particular website.
-     * @return {boolean} whether the session was retired.
+     * @returns Whether the session was retired.
      */
     retireOnBlockedStatusCodes(statusCode: number, blockedStatusCodes: number[] = []): boolean {
         const isBlocked = STATUS_CODES_BLOCKED.concat(blockedStatusCodes).includes(statusCode);
@@ -332,7 +330,7 @@ export class Session {
      * Returns cookies saved with the session in the typical
      * key1=value1; key2=value2 format, ready to be used in
      * a cookie header or elsewhere.
-     * @return {string} represents `Cookie` header.
+     * @returns Represents `Cookie` header.
      */
     getCookieString(url: string): string {
         return this.cookieJar.getCookieStringSync(url, {});
@@ -341,7 +339,6 @@ export class Session {
     /**
      * Transforms puppeteer cookie to tough-cookie.
      * @param puppeteerCookie Cookie from puppeteer `page.cookies` method.
-     * @internal
      */
     protected _puppeteerCookieToTough(puppeteerCookie: PuppeteerCookie): Cookie {
         const isExpiresValid = puppeteerCookie.expires && typeof puppeteerCookie.expires === 'number' && puppeteerCookie.expires > 0;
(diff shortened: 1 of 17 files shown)


vladfrangu:
> So I could basically use the DatasetDataOptions interface there, but in order to hide some params, I am using DatasetIteratorOptions extends DatasetDataOptions with some params overridden with @internal.

Ideally you could make a base interface with common props, then extend it as needed, or use Omit to remove properties that aren't needed

packages/apify/src/storages/key_value_store.ts
```diff
@@ -205,18 +205,15 @@ export class KeyValueStore {
...
-    // TODO: type options
-    async setValue<T>(key: string, value: T | null, options: Dictionary = {}): Promise<void> {
+    async setValue<T>(key: string, value: T | null, options: RecordOptions = {}): Promise<void> {
         ow(key, 'key', ow.string.nonEmpty);
         ow(key, ow.string.validate((k) => ({
             validator: ow.isValid(k, ow.string.matches(KEY_VALUE_STORE_KEY_REGEX)),
             message: 'The "key" argument must be at most 256 characters long and only contain the following characters: a-zA-Z0-9!-_.\'()',
         })));
-        if (options.contentType && !ow.isValid(value, ow.any(ow.string, ow.buffer))) {
+        if (options!.contentType && !ow.isValid(value, ow.any(ow.string, ow.buffer))) {
```
vladfrangu, line 216: is the `!` needed?

packages/apify/src/storages/key_value_store.ts
```diff
@@ -436,11 +405,9 @@ export async function getValue<T extends Dictionary | string | Buffer>(key: stri
...
- * @param [options]
- * @param {string} [options.contentType]
- *   Specifies a custom MIME content type of the record.
+ * @param [options] Record options.
  */
-export async function setValue<T>(key: string, value: T, options?: unknown): Promise<void> {
+export async function setValue<T>(key: string, value: T, options: RecordOptions = {}): Promise<void> {
     const store = await openKeyValueStore();
 
     // @ts-ignore options need to be typed
```
vladfrangu, line 413: This isn't needed anymore, right?

packages/apify/src/storages/key_value_store.ts
```diff
@@ -498,3 +465,17 @@ export interface KeyValueStoreOptions {
     client: ApifyClient | ApifyStorageLocal;
     isLocal?: boolean;
 }
+
+interface RecordOptions {
```
vladfrangu, line 469: Please export the option interfaces

vladfrangu (changes requested):
Awesome start! Just a few small nits from me

AndreyBykov:
@vladfrangu but that's the thing, there's `getData` call (which takes all options) withing `forEach` (here we expose only 3 props). So I guess they should not be omitted at all, but rather be internal in this case...

packages/apify/src/storages/request_queue.ts
```diff
@@ -596,7 +593,7 @@ export class RequestQueue {
      * Adds a request straight to the queueHeadDict, to improve performance.
      * @private
      */
-    _maybeAddRequestToQueueHead(requestId, forefront) {
+    _maybeAddRequestToQueueHead(requestId: string, forefront: boolean): void {
```
B4nan, line 596: is this used from outside? if not, make it `private` on TS level

packages/apify/src/storages/request_queue.ts
```diff
@@ -676,9 +669,10 @@ export class RequestQueue {
      *   ID or name of the request queue to be opened. If `null` or `undefined`,
      *   the function returns the default request queue associated with the actor run.
      * @param [options]
-     * @param [options.forceCloud=false]
+     * @param [options.forceCloud]
      *   If set to `true` then the function uses cloud storage usage even if the `APIFY_LOCAL_STORAGE_DIR`
      *   environment variable is set. This way it is possible to combine local and cloud storage.
+     * @default false
      */
     static async open(queueIdOrName?: string | null, options: { forceCloud?: boolean } = {}): Promise<RequestQueue> {
```
B4nan, line 677: let's make this an interface too, i don't think the API docs can handle the `options.forceCloud` jsdoc

test/storages/request_queue.test.ts
```diff
@@ -512,13 +513,13 @@ describe('RequestQueue remote', () => {
         const updateRequestMock = jest.spyOn(queue.client, 'updateRequest');
         updateRequestMock.mockResolvedValueOnce({ requestId: 'b', wasAlreadyHandled: false, wasAlreadyPresent: true });
 
-        await queue.markRequestHandled(requestBWithId);
+        await queue.markRequestHandled(requestBWithId as ApifyRequest);
```
B4nan, line 516: let's move the type assertion to the place where the variable is defined

B4nan (changes requested):
looking good, few more nits from my end

packages/apify/src/storages/dataset.ts
```diff
@@ -85,6 +85,81 @@ export function chunkBySize(items: string[], limitBytes: number): string[] {
...
+    /**
+     * If `true` then the function doesn't return empty items.
+     * Note that in this case the returned number of items might be lower than limit parameter and pagination must be done using the `limit` value.
+     * @default false
+     */
+    skipEmpty?: boolean;
+}
+
+interface DatasetIteratorOptions extends DatasetDataOptions {
+    /** @internal */
```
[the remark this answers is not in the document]
AndreyBykov: But looking at the current docs - only 3 properties are exposed in `.forEach` and `.map` methods. I thought if I would just use the main interface (without hidden/internal props) - we would show all the props for these methods? Is it kinda ok?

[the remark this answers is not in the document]
B4nan:
```ts
export interface DatasetIteratorOptions extends Omit<DatasetDataOptions, 'offset' | 'limit' ...> {
```

packages/apify/src/storages/key_value_store.ts
```diff
@@ -281,26 +273,16 @@ export class KeyValueStore {
...
-     * @param {Record<string, any>} [options]
-     * @param {number} [index=0]
-     * @return {Promise<Promise<void> | undefined>}
-     * @private
-     */
-    private async _forEachKey(iteratee: KeyConsumer, options: Dictionary<string | undefined> = {}, index = 0): Promise<void> {
-        const { exclusiveStartKey } = options;
+    /** @private */
+    private async _forEachKey(iteratee: KeyConsumer, options: KeyValueStoreIteratorOptions = {}, index = 0): Promise<void> {
+        const { exclusiveStartKey } = options!;
```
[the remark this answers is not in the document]
AndreyBykov: yeah, I removed the default value at the beginning - TS was not happy, then I realized it should be there and forgot to remove the `!`

packages/apify/src/storages/request_queue.ts
```diff
@@ -668,13 +672,10 @@ export class RequestQueue {
      * @param [queueIdOrName]
      *   ID or name of the request queue to be opened. If `null` or `undefined`,
      *   the function returns the default request queue associated with the actor run.
-     * @param [options]
-     * @param [options.forceCloud]
-     *   If set to `true` then the function uses cloud storage usage even if the `APIFY_LOCAL_STORAGE_DIR`
-     *   environment variable is set. This way it is possible to combine local and cloud storage.
+     * @param [options] Open Request Queue options.
      * @default false
```
B4nan, line 676: this belongs to the `forceCloud`, right?

B4nan (approved):

packages/apify/src/storages/dataset.ts
```diff
@@ -85,6 +85,81 @@ export function chunkBySize(items: string[], limitBytes: number): string[] {
...
+     * If `true` then the function doesn't return empty items.
+     * Note that in this case the returned number of items might be lower than limit parameter and pagination must be done using the `limit` value.
+     * @default false
+     */
+    skipEmpty?: boolean;
+}
+
+export interface DatasetIteratorOptions extends Omit<DatasetDataOptions, 'offset' | 'limit' | 'clean' | 'skipHidden' | 'skipEmpty'> {
+    /** @internal */
+    offset?: number;
```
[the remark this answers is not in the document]
AndreyBykov: Yeah, we had a call with Martin. It's more about docs. These fields are further used inside of the method, but we do not expose them, so not to have any clashes (public/internal) - I omit and re-define them there with @internal.

packages/apify/src/storages/request_queue.ts
```diff
@@ -500,21 +504,19 @@ export class RequestQueue {
...
-     * @param {boolean} [ensureConsistency=false] If true then query for queue head is retried until queueModifiedAt
+     * @param [ensureConsistency] If true then query for queue head is retried until queueModifiedAt
      *   is older than queryStartedAt by at least API_PROCESSED_REQUESTS_DELAY_MILLIS to ensure that queue
      *   head is consistent.
-     * @param {number} [limit] How many queue head items will be fetched.
-     * @param {number} [iteration] Used when this function is called recursively to limit the recursion.
+     * @default false
+     * @param [limit] How many queue head items will be fetched.
+     * @param [iteration] Used when this function is called recursively to limit the recursion.
      * @return {Promise<boolean>} Indicates if queue head is consistent (true) or inconsistent (false).
```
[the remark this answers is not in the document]
AndreyBykov: Yeah, think so. Will go through everything once again, thanks for noticing!

AndreyBykov:
I went through all folders once again, could you guys check the last commit?
@vladfrangu - removed the return types.
@B4nan - cleaned up the @internal for protected/private methods.
Wanted to include both in this PR since we discussed it before in comments.

B4nan (approved):

packages/apify/src/session_pool/session.ts
```diff
@@ -222,11 +218,13 @@ export class Session {
...
     /**
      * Gets session state for persistence in KeyValueStore.
-     * @return {SessionState} represents session internal state.
+     * @returns Represents session internal state.
      */
-    getState() {
+    getState(): SessionState {
         return {
             id: this.id,
+            // @ts-expect-error Type 'Serialized' is missing the following properties from type 'CookieJar':
```
vladfrangu, line 226: This doesn't seem right, is cookieJar meant to be an object or an actual instance?

vladfrangu (approved):
LGTM, just one question for now

packages/apify/src/session_pool/session.ts
AndreyBykov: @vladfrangu I think here it's supposed to be the serialized. Session in memory keeps the actual instance of CookieJar. But this method gets the session state to be persisted in KeyValueStore and some fields are omitted (e.g. setCookie, setCookieSync methods) - thus the error.

PR #12, opened 2021-12-06
chore: add more TS types (last batch)

AndreyBykov:
Last batch of types updates.

1. I skipped `utils_log.ts`, it's mentioned that it's for building docs mainly. Should I update `types-apify/apify-shared/log.d.ts` instead, or I could just skip it indeed?
2. Tests should be passing, but I would ask you to kinda double check the `configuration.ts` and `request.ts` (`pushErrorMessage` specifically).
3. `utils_request.ts` - I am kinda confused about this one - there seems to be some mess with types (it's not finished). Should I have `GotScrapingOptions` as an interface here? There's also some mismatch between `stream.Duplex` vs `CancelableRequest` (there was also a TODO about it). Could you point me to what's kinda the right approach here or should I also skip it for now?
4. `request.test.ts` - there's one test failing on `request.pushErrorMessage`. The test is pushing different types to the errors array, while the method expects Error/string/Object. I just don't kinda get it - why is the test done this way? Why random stuff like functions etc should be pushed there? Aren't we explicitly mean in the method signature that it should not be function/symbol/etc?

And well - of course - waiting for other comments. Thanks in advance!
And yeah - sry, I kinda expected this PR to be tiny, but it's not in the end :/

### DIFF
--- a/packages/apify/src/apify.ts
+++ b/packages/apify/src/apify.ts
@@ -1,11 +1,11 @@
 import ow from 'ow';
 import { ACT_JOB_STATUSES, ENV_VARS } from '@apify/consts';
-import { ApifyClient } from 'apify-client';
+import { ApifyClient, ApifyClientOptions } from 'apify-client';
 import { WebhookOptions, CallOptions, CallTaskOptions, getEnv, UserFunc, WebhookRun, ApifyEnv } from './actor';
 import { initializeEvents, stopEvents } from './events';
-import { StorageManager } from './storages/storage_manager';
+import { StorageManager, StorageManagerOptions } from './storages/storage_manager';
 import { Dataset } from './storages/dataset';
-import { KeyValueStore, maybeStringify } from './storages/key_value_store';
+import { KeyValueStore, RecordOptions, maybeStringify } from './storages/key_value_store';
 import { RequestList, RequestListOptions, REQUESTS_PERSISTENCE_KEY, STATE_PERSISTENCE_KEY } from './request_list';
 import { RequestQueue } from './storages/request_queue';
 import { SessionPool, SessionPoolOptions } from './session_pool/session_pool';
@@ -238,7 +238,7 @@ export class Apify {
         callActorOpts.token = token;
 
         if (input) {
-            callActorOpts.contentType = addCharsetToContentType(callActorOpts.contentType);
+            callActorOpts.contentType = addCharsetToContentType(callActorOpts.contentType!);
             input = maybeStringify(input, callActorOpts);
         }
 
@@ -305,7 +305,7 @@ export class Apify {
      *  Input overrides for the actor task. If it is an object, it will be stringified to
      *  JSON and its content type set to `application/json; charset=utf-8`.
      *  Provided input will be merged with actor task input.
-     * @param {object} [options]
+     * @param [options]
      * @throws {ApifyCallError} If the run did not succeed, e.g. if it failed or timed out.
      */
     async callTask(taskId: string, input?: Dictionary | string, options: CallTaskOptions = {}): Promise<ActorRun> {
@@ -379,8 +379,8 @@ export class Apify {
      * @param [input]
      *  Input for the actor. If it is an object, it will be stringified to
      *  JSON and its content type set to `application/json; charset=utf-8`.
-     *  Otherwise the `options.contentType` parameter must be provided.
-     * @param {object} [options]
+     *  Otherwise, the `options.contentType` parameter must be provided.
+     * @param [options]
      */
     async metamorph(targetActorId: string, input?: Dictionary | string, options: MetamorphOptions = {}): Promise<void> {
         ow(targetActorId, ow.string);
@@ -402,7 +402,7 @@ export class Apify {
         if (!runId) throw new Error(`Environment variable ${ENV_VARS.ACTOR_RUN_ID} must be provided!`);
 
         if (input) {
-            metamorphOpts.contentType = addCharsetToContentType(metamorphOpts.contentType);
+            metamorphOpts.contentType = addCharsetToContentType(metamorphOpts.contentType!);
             input = maybeStringify(input, metamorphOpts);
         }
 
@@ -422,7 +422,7 @@ export class Apify {
      * In local environment, the function will print a warning and have no effect.
      *
      * @param options
-     * @return {Promise<WebhookRun | undefined>} The return value is the Webhook object.
+     * @returns The return value is the Webhook object.
      * For more information, see the [Get webhook](https://apify.com/docs/api/v2#/reference/webhooks/webhook-object/get-webhook) API endpoint.
      */
     async addWebhook(options: WebhookOptions): Promise<WebhookRun | undefined> {
@@ -499,11 +499,8 @@ export class Apify {
      *   ID or name of the dataset to be opened. If `null` or `undefined`,
      *   the function returns the default dataset associated with the actor run.
      * @param [options]
-     * @param [options.forceCloud=false]
-     *   If set to `true` then the function uses cloud storage usage even if the `APIFY_LOCAL_STORAGE_DIR`
-     *   environment variable is set. This way it is possible to combine local and cloud storage.
      */
-    async openDataset(datasetIdOrName?: string | null, options: { forceCloud?: boolean } = {}): Promise<Dataset> {
+    async openDataset(datasetIdOrName?: string | null, options: Omit<StorageManagerOptions, 'config'> = {}): Promise<Dataset> {
         ow(datasetIdOrName, ow.optional.string);
         ow(options, ow.object.exactShape({
             forceCloud: ow.optional.boolean,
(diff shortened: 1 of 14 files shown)


B4nan:
yeah i was hoping this will be done before we start with other refactors, lets merge it there - changed the target already, @AndreyBykov please fix those conflicts

AndreyBykov:
Some conflicts are actually exactly about some things I mentioned above. E.g. - do we need `CreateClientOptions` (and similar) interfaces while it's basically `ApifyClientOptions` with some options omitted - should I use separate interface for this, or use Omit<ApifyClientOptions, ...>? Otherwise I think it's fine, will fix in a few minutes.

B4nan:
> Some conflicts are actually exactly about some things I mentioned above. E.g. - do we need CreateClientOptions (and similar) interfaces while it's basically ApifyClientOptions with some options omitted - should I use separate interface for this, or use Omit<ApifyClientOptions, ...>? Otherwise I think it's fine, will fix in a few minutes.

Maybe we don't even need to omit anything (as long as we are passing down the whole options object).

packages/apify/src/apify.ts
```diff
@@ -532,15 +529,14 @@ export class Apify {
...
-     * @returns {Promise<Object<string, *>|string|Buffer|null>}
+     * @param key Unique record key.
+     * @returns
      *   Returns a promise that resolves to an object, string
      *   or [`Buffer`](https://nodejs.org/api/buffer.html), depending
      *   on the MIME content type of the record, or `null`
      *   if the record is missing.
      */
-    async getValue(key) {
+    async getValue(key: string): Promise<Record<string, unknown> | string | Buffer | null> {
```
B4nan, line 539: we should add generics here, cc @vladfrangu

packages/apify/src/apify.ts
```diff
@@ -604,13 +598,13 @@ export class Apify {
...
      *
-     * @returns {Promise<Object<string, *>|string|Buffer|null>}
+     * @returns
      *   Returns a promise that resolves to an object, string
      *   or [`Buffer`](https://nodejs.org/api/buffer.html), depending
      *   on the MIME content type of the record, or `null`
      *   if the record is missing.
      */
-    async getInput() {
+    async getInput(): Promise<Record<string, unknown> | string | Buffer | null> {
```
B4nan, line 607: same here

packages/apify/src/apify.ts
```diff
@@ -823,26 +811,19 @@ export class Apify {
...
-     *
-     * @returns {boolean}
      */
-    isAtHome() {
+    isAtHome(): boolean {
         return !!this.config.get('isAtHome');
     }
 
-    get utils() {
+    get utils(): Record<string, unknown> {
```
B4nan, line 826: this needs to be properly typed, we need to know the exact keys as well as their values

on the other hand iirc we are planning to ditch this static utils register, so maybe its not worth it? cc @mnmkng

packages/apify/src/utils.ts
```diff
@@ -80,13 +80,8 @@ const psTreePromised = util.promisify(psTree);
...
  * environment variables. You can override the token via the available options. That's useful
  * if you want to use the client as a different Apify user than the SDK internals are using.
- *
- * @param {object} [options]
- * @param {string} [options.token]
- * @param {string} [options.maxRetries]
- * @param {string} [options.minDelayBetweenRetriesMillis]
  */
-export function newClient(options = {}): ApifyClient {
+export function newClient(options: Omit<ApifyClientOptions, 'baseUrl' | 'requestInterceptors' | 'timeoutSecs'> = {}): ApifyClient {
```
B4nan, line 84: maybe we should use the whole `ApifyClientOptions`? `baseUrl` is definitely supported, and given what this method is used for, instead of omitting props from there, i would loosen the `ow` validation to support everything from that interface

packages/apify/src/utils_request.ts
```diff
@@ -383,11 +384,10 @@ function ensureRequestIsDispatched(duplexStream, gotScrapingOptions) {
 }
 
 /**
- * @param {RequestAsBrowserOptions} options
  * @ignore
  * @internal
```
B4nan, line 388: one annotation is enough, i would maybe go with `@internal` but it does not really matter

test/request.test.ts
```diff
@@ -87,16 +87,23 @@ describe('Apify.Request', () => {
         };
 
         request.pushErrorMessage(undefined);
+        // @ts-expect-error Argument of type 'boolean' is not assignable to parameter of type 'string | Error'.
```
B4nan, line 90: if those calls do not produce runtime error and are technically supported, maybe we should adjust the method signature to allow them?

maybe we should use just `unknown` in there

B4nan (changes requested):

packages/apify/src/apify.ts
mnmkng: Yeah, let's ignore it for now. I'm really looking forward to having the utils available directly from the `crawlingContext`. E.g.

```js
function handlePageFunction(ctx) {
    // ...
    await ctx.enqueueLinks({ selector: 'a.item' })
}
```

Instead of the current:

```js
function handlePageFunction(ctx) {
    // ...
    await Apify.utils.enqueueLinks({
        page: ctx.page,
        requestQueue: ctx.crawler.requestQueue,
        selector: 'a.item'
    })
}
```

packages/apify/src/utils.ts
AndreyBykov: The point here is again - what we expose to the docs and what we do not. Previously it was only certain props, but yeah, originally I used the whole `ApifyClientOptions`. Then I just thought I would keep it the same. Kinda the same with some conflicts that I resolved yesterday - I could just use the whole `ApifyClientOptions` interface there (now I opted for `CreatClientOptions` added by Vlad).

B4nan: Yeah I know, but I dont see any reason why we should not expose everything here, `newClient` is basically just a shortcut for `new ApifyClient`, so why not allow passing all the options that the class accepts. There is already a validation in the constructor, so I would be ok with dropping it from here too.

AndreyBykov: aha, cool, got it

packages/apify/src/apify.ts
vladfrangu: We can probably copy the generics from the KeyValue store and put them here as well, right?

B4nan: yop

B4nan:
@AndreyBykov dont forget to deal with the conflicts

image

AndreyBykov:
@B4nan of course, I am finishing with fixes, then conflicts, then we'll have a call with @vladfrangu and then will merge it

packages/apify/src/apify.ts
```diff
@@ -305,7 +305,7 @@ export class Apify {
      *  Input overrides for the actor task. If it is an object, it will be stringified to
      *  JSON and its content type set to `application/json; charset=utf-8`.
      *  Provided input will be merged with actor task input.
-     * @param {object} [options]
+     * @param [options]
      * @throws {ApifyCallError} If the run did not succeed, e.g. if it failed or timed out.
      */
     async callTask(taskId: string, input?: Dictionary | Dictionary[], options: CallTaskOptions = {}): Promise<ActorRunWithOutput> {
```
vladfrangu, line 311: I know it's technically unrelated to this PR, but should we use the options from apify-client instead? CC @B4nan

packages/apify/src/utils_request.ts
```diff
@@ -243,7 +243,8 @@ export async function requestAsBrowser<T = string>(options: RequestAsBrowserOpti
                     return reject(e);
                 }
 
-                addResponsePropertiesToStream(duplexStream, res);
+                // TODO: stream vs response interface?
+                addResponsePropertiesToStream(duplexStream as any, res);
```
vladfrangu, line 247: This seems to be a merge/rebase mistake, as in the branch this cast isn't needed

vladfrangu (changes requested):
Looks great! Just two small things, otherwise LGTM from me

packages/apify/src/apify.ts
AndreyBykov: @vladfrangu should I do it here or you will update it? I see some options are not exactly the same (e.g. `memory` in actor/task start options, while here it's `memoryMbytes`. Same with `timeout` vs `timeoutSecs`, `waitSecs` vs `waitForFinish, `webhooks` are of the different type. Should I rename props too or extend it and keep the same names?

vladfrangu: Hello fellow night owl! 🦉
I'd say we all should have a call about this (especially the memory renames) and decide how we want to move forward. Does that sound good?

AndreyBykov: :wave: 🦉 :smile:
Yep, I removed the unnecessary typecasting (your second comment). So I guess this point is the last thing to agree on before merging if @B4nan would not have any other comments. So let's have a call tomorrow.

vladfrangu (approved):

B4nan (approved):

PR #17, opened 2022-01-13
refactor: deal with some TODOs, improve some types, make some BCs

B4nan:


mnmkng (approved):

packages/apify/src/request_list.ts
```diff
@@ -244,23 +243,21 @@ export class RequestList {
     /**
      * Dictionary of requests that were returned by fetchNextRequest().
      * The key is uniqueKey, value is true.
-     * TODO: Change this to Set
-     * @internal (TODO or should be public maybe?)
+     * @internal
      */
-    inProgress: Record<string, boolean> = {};
+    inProgress = new Set<string>();
```
mnmkng, line 248: The comments do not reflect reality anymore.

packages/apify/src/request_list.ts
```diff
@@ -244,23 +243,21 @@ export class RequestList {
...
     /**
      * Dictionary of requests for which reclaimRequest() was called.
      * The key is uniqueKey, value is true.
      * Note that reclaimedRequests is always a subset of inProgress!
-     * TODO: Change this to Set
-     * @internal (TODO or should be public maybe?)
+     * @internal
      */
-    reclaimed: Record<string, boolean> = {};
+    reclaimed = new Set<string>();
```
mnmkng, line 256: ditto

packages/apify/src/request_list.ts
```diff
@@ -244,23 +243,21 @@ export class RequestList {
...
-     * @internal (TODO or should be public maybe?)
+     * @internal
      */
-    inProgress: Record<string, boolean> = {};
+    inProgress = new Set<string>();
 
     /**
      * Dictionary of requests for which reclaimRequest() was called.
      * The key is uniqueKey, value is true.
      * Note that reclaimedRequests is always a subset of inProgress!
```
mnmkng, line 253: TBH I don't get this `Note that reclaimedRequests is always a subset of inProgress!`.

It seems that `inProgress` should always have all that were ever picked for usage. And then reclaimed duplicates them, if they were scraped at least once. It's kinda weird, but I guess it's because this is not a queue. The index only ever goes up, so requests must never be removed from inProgress until they are handled.

B4nan: Let's remove that, it does not make sense when we have sets. With dicts it was probably true as we just swapped the true/false value, now we literally delete such element from the set.

PR #18, opened 2022-01-17
chore: new homepage

### LINKED ISSUE #13: Homepage redesign
- https://www.figma.com/file/LEscoqX1kg6IUuyidPoO6n/SDK-homepage?node-id=0%3A1 
- no need to be pixel perfect
- mainly fonts and sizes, links/buttons style, images, backgrounds, general layout and text/content
- we want to keep the dark theme switcher, search box, both on the right side as they are now
- ensure responsiveness

szmarczak:
Fixes #13

 - [x] light mode
 - [x] dark mode
 - [x] responsiveness

it's almost 1:1

[Screenshot](https://user-images.githubusercontent.com/36894700/149714122-4f291d42-6b3b-461d-9d0a-0a1fa90321a9.png)


website/docusaurus.config.js
```diff
@@ -104,22 +105,12 @@ module.exports = {
...
-                    label: versions[0],
-                    position: 'left',
-                    items: [
-                        { label: 'Next', to: 'api/next' },
-                        ...versions.map((version, i) => ({
-                            label: version,
-                            to: i === 0 ? 'api' : `api/${version}`,
-                            exact: true,
-                        })),
-                    ],
```
B4nan, line 118: this has to stay for now, we cant reuse the native version picker for API docs unfortunately, so we need both (and we will need to find a way to toggle between them dynamically so only one is visible most probably)

B4nan (approved):

website/docusaurus.config.js
szmarczak: > this has to stay for now, we cant reuse the native version picker for API docs unfortunately

Why?

B4nan: Because it does not respect the api docs at all

PR #20, opened 2022-01-17
chore: migrating actor-scraper to monorepo

vladfrangu:


packages/scraper-tools/package.json
```diff
@@ -0,0 +1,47 @@
...
+        "ajv": "^8.8.2",
+        "content-type": "^1.0.4"
+    },
+    "peerDependencies": {
+        "apify": ">=2.x || >=2.1.1-beta.0"
+    },
+    "devDependencies": {
+        "@types/fs-extra": "^9.0.13",
+        "@types/puppeteer": "^5.4.4",
+        "apify": "file:../apify",
```
B4nan, line 44: you dont need dev deps on internal packages, they are available already due to TS paths mapping configuration

(just one think i noticed on the first look, will check this out locally before i give you more feedback)

packages/scraper-tools/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
B4nan, line 1: also we dont need licenses in packages, there is on in the root of the repo that gets copied to each package during build (thats the `copy.ts` script)

package.json
```diff
@@ -45,7 +46,7 @@
         "postinstall": "npx husky install",
         "prepublishOnly": "npx lerna run copy",
         "clean": "lerna run clean",
-        "build": "lerna run build",
+        "build": "lerna run build --sort",
```
B4nan, line 49: this should also not be needed, and I really do believe this is enabled by default (the option is there so you can use it as `--no-sort` to disable things)

packages/actor-scraper/cheerio-scraper/package.json
```diff
@@ -0,0 +1,37 @@
+{
+	"name": "actor-cheerio-scraper",
+	"version": "2.0.0",
+	"description": "Crawl web pages using HTTP requests and Cheerio",
+	"main": "main.js",
+	"dependencies": {
+		"@apify/scraper-tools": "^0.4.4",
```
B4nan, line 7: if this is now package in monorepo, we should align its version with other packages

packages/apify/src/proxy_configuration.ts
```diff
@@ -55,6 +55,8 @@ export interface ProxyConfigurationOptions {
      */
     apifyProxyCountry?: string;
 
+    useApifyProxy?: boolean;
```
B4nan, line 58: remove those changes (including the one on line 528), it should not be public, its an internal hack

packages/actor-scraper/web-scraper/tsconfig.json
```diff
@@ -0,0 +1,13 @@
+{
+	"extends": "@apify/tsconfig",
+	"compilerOptions": {
+		"outDir": "dist",
+		"module": "ESNext",
```
B4nan, line 5: do we really build into esnext? rest of the packages won't be build that way for sure, we need to stick to CJS for now

packages/actor-scraper/web-scraper/src/internals/global_store.ts
```diff
@@ -0,0 +1,35 @@
...
+    override set(key: string, value: V) {
+        if (typeof key !== 'string') throw new Error('GlobalStore#set parameter "key" must be a string.');
+        return super.set(key, value);
+    }
+
+    override forEach(): never {
+        throw new Error('GlobalStore#forEach function is not available due to underlying technology limitations.');
+    }
+
+    // @ts-expect-error Overriding return type
```
B4nan, line 21: seems better to cast the return value as `any`, we should not use ts-expect-error for things that can be casted easily, i still do consider that as a code smell

packages/actor-scraper/web-scraper/src/internals/bundle.browser.ts
```diff
@@ -0,0 +1,235 @@
+/* eslint-disable max-classes-per-file */
+/* istanbul ignore next */
+
+// TODO: properly type this file, mostly copy paste from context.ts in scraper-tools
```
B4nan, line 4: i believe this file was done in a way so we dont need to compile it anyhow. if we want to have it in TS, we might need some additional things to build it properly (all packages are for node 15+, we cant apply the same constraints to this file as its for browsers)

packages/actor-scraper/puppeteer-scraper/.gitignore
```diff
@@ -0,0 +1,7 @@
+.idea
```
B4nan, line 1: same here, should be removed

packages/actor-scraper/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "actor-scraper",
+  "version": "0.0.1",
+  "description": "This repository houses all of Apify generic actors that are used for simplified scraping using a pre-defined, schema validated UI input instead of the typical JSON input used in other actors.",
+  "private": true,
+  "workspaces": [
```
B4nan, line 6: this seems wrong too, should be part only of the root package.json

tsconfig.eslint.json
```diff
@@ -1,8 +1,8 @@
 {
 	"extends": "./tsconfig.json",
 	"include": [
-		"packages/*/src/**/*.js",
-		"packages/*/src/**/*.ts",
+		"packages/**/src/**/*.js",
+		"packages/**/src/**/*.ts",
```
B4nan, line 5: this should go away too, we dont want nested packages

B4nan (changes requested):
havent seen everything properly and its still WIP anyway, so just few notes about things i caught :]

would be great to finish this during the week, i was hoping to have it merged already as it kinda blocks the E2E tests

package.json
vladfrangu: It seems that making scraper-tools depend on the local apify version made lerna run the build scripts in the right order, so I will remove this

packages/actor-scraper/web-scraper/test/.eslintrc.json
```diff
@@ -0,0 +1,16 @@
+{
```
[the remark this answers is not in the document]
vladfrangu: I didn't get to remove this file, will remove

package.json
```diff
@@ -9,7 +9,8 @@
     "main": "build/index.js",
     "types": "build/index.d.ts",
     "workspaces": [
-        "packages/*"
+        "packages/*",
+        "packages/actor-scraper/*"
```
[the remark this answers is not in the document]
vladfrangu: I can remove it, but then how will dependencies be managed in the actors? Just run `npm i` in each of them? We can do that I suppose..

packages/actor-scraper/web-scraper/tsconfig.json
vladfrangu: I copied this from the web checker actor I wrote, which runs on ESM, I'll remove this 👍

package.json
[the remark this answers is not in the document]
B4nan: The purpose of this PR should be just to move them, we can think about this later. Actors wont be published to NPM and they won't be tested the same way as regular packages (we will have the E2E test suite for that), so I can imagine this is not really a problem. But maybe we will move them to the top level too, let's see in following weeks. 

I am not totally against this, if it works as expected, let's keep it for now.

packages/actor-scraper/web-scraper/tsconfig.json
B4nan: As mentioned in the other comment, let's keep it here for now, those are not regular packages so it might be just fine. I can imagine we will have a separate build process for actors so it might not matter at all.

B4nan:
just a motivation for you to finish this sooner :P i am almost done with the initial monorepo split (actor sdk vs core package), which will be quite breaking for this (all imports will need to be adjusted, and there are more smaller BCs in public API too)

its again quite huge :D ` 132 files changed, 2881 insertions(+), 3182 deletions(-)`

vladfrangu:
Oh jeez... That'll be a fun PR to review 😅

B4nan:
I guess I will cheat a bit and merge it when the CI is green :D But of course will be more than happy to any reviewers :] Still have some tests failing, but its less than 10 right now :]

vladfrangu:
Hmm, I found a bit of an issue with making the actors ESM, that is that our modules don't have ESM exports available. Do we want to re-use the `gen-esm-wrapper` module to generate those or something like tsup?

PR #22, opened 2022-01-24
refactor: separate crawler related code to core package

B4nan:


jest.config.js
```diff
@@ -13,14 +13,20 @@ module.exports = {
...
+        '^@crawlers/(.*)$': '<rootDir>/packages/$1/src',
     },
     modulePathIgnorePatterns: [
         'dist/package.json',
         '<rootDir>/package.json',
     ],
     globals: {
         'ts-jest': {
-            tsconfig: 'test/tsconfig.json',
+            tsconfig: 'test/apify/tsconfig.json',
```
vladfrangu, line 29: We should probably move the tsconfig to the test folder instead

packages/apify/src/actor.ts
```diff
@@ -1,458 +1,1320 @@
...
+        if (!this.isAtHome()) {
+            log.warning('Actor.metamorph() is only supported when running on the Apify platform.');
+            return;
+        }
+
+        const {
+            customAfterSleepMillis = this.config.get('metamorphAfterSleepMillis'),
+            ...metamorphOpts
+        } = options;
+        const runId = this.config.get('actorRunId')!;
```
vladfrangu, line 301: Probably a nit for the future but we should probably have an object/enum for these keys, so we can strictly type them

packages/apify/src/actor.ts
```diff
@@ -1,458 +1,1320 @@
...
+        await this.newClient().run(runId).metamorph(targetActorId, input, metamorphOpts);
```
vladfrangu, line 302: Shouldn't you reuse this.apifyClient instead?

packages/apify/src/actor.ts
```diff
@@ -1,458 +1,1320 @@
...
+        }));
+
+        const { eventTypes, requestUrl, payloadTemplate, idempotencyKey } = options;
+
+        if (!this.isAtHome()) {
+            log.warning('Actor.addWebhook() is only supported when running on the Apify platform. The webhook will not be invoked.');
+            return undefined;
+        }
+
+        const runId = process.env[ENV_VARS.ACTOR_RUN_ID];
```
vladfrangu, line 334: Shouldn't this use this.config.get, for consistency?

vladfrangu (changes requested):

packages/apify/src/actor.ts
B4nan: what you mean? config is already strictly typed, it both checks the keys as well as maps the return type based on it

packages/apify/src/actor.ts
B4nan: i believe there was some reason originally, but now its all rewritten so it should preserve the old behaviour from static helpers

so yes, i guess it should be safe to use it here (tests do pass)

vladfrangu (approved):

PR #23, opened 2022-01-25
feat: bring some turbo in the monorepo

vladfrangu:


B4nan (changes requested):
not sure if we really need to lint on each node version we test, but lets keep it there and see how it works. also i though we dont need to build the project to lint it (but the build is cached, so it shouldnt matter i guess?)

vladfrangu:
Hmm, thats a good catch, we shouldn't lint on all node versions. As for building on lint, technically it shouldn't be needed (and I can remove it), but we have encountered issues with typescript eslint rules that broke because of co-dependent packages (your call).

B4nan:
I like to deal with issues after I see them, never had such problem myself :] Let's remove that, we can easily add it back if we see its needed.

PR #25, opened 2022-01-28
feat: add public actor scrapers e2e tests

AndreyBykov:
Added the tests currently present on the platform. There are some things I wanted to discuss/confirm:
1) Web scraper dev mode - it needs APIFY_CONTAINER_URL and APIFY_CONTAINER_PORT to start with dev mode on - I added localhost:8000 in order for the test to actually work. But locally there is kinda no container. Should we leave it like this? I am thinking maybe to add a condition that it will check whether test is running locally or not. Or maybe even simple `process.env.APIFY_CONTAINER_PORT = process.env.APIFY_CONTAINER_PORT || '8000';` would work, what do you think?
2) Web scraper proxy rotation - it needs access to the proxy (i.e. Apify token/proxy password). How do you think it's better to handle it here? Just add the password/token for each `npm run test:e2e` call? As we could not simply hardcode it there.
3) ignore ssl errors enabled/disabled - for web scraper and puppeteer it works for different number of pages (when ignored -there are more successfully finished requests), while for cheerio this number is the same in both cases. Same on the platform. So basically test works the same locally, but I don't think it should be this way generally speaking, or am I missing something related to got-scraping?
4) Cheerio scraper xml content type - there's a test adding `additionalMimeTypes: ['text/xml']`, but I think it works the same with/without it - so this test does not make any sense really. I guess 'text/xml' type was added at some point as the type supported by default (not 100% about that). But if that's the case - we should probably ditch this test (or update it t some other content-type).

Otherwise all of them finish successfully.


AndreyBykov:
@B4nan 1 and 2 are not relevant anymore (double-checked - tests pass, so loading the token from local auth file works just fine).

test/e2e/run.mjs
```diff
@@ -2,12 +2,15 @@ import { dirname } from 'node:path';
...
 import { isMainThread, Worker, workerData } from 'worker_threads';
-import { colors } from './tools.mjs';
+import { colors, getApifyToken } from './tools.mjs';
 
 const basePath = dirname(fileURLToPath(import.meta.url));
 
 process.env.APIFY_LOG_LEVEL = 0; // switch off logs for better test results visibility
 process.env.APIFY_HEADLESS = 1; // run browser in headless mode (default on platform)
+process.env.APIFY_TOKEN = process.env.APIFY_TOKEN ?? await getApifyToken();
+process.env.APIFY_CONTAINER_URL = process.env.APIFY_CONTAINER_URL ?? 'https://localhost';
```
B4nan, line 12: maybe `http://localhost`? not sure if it actually matters, but not many people do have https set up locally

and i would maybe prefer `127.0.0.1`

test/e2e/tools.mjs
```diff
@@ -6,6 +6,7 @@ import { fileURLToPath } from 'node:url';
 import { setTimeout } from 'node:timers/promises';
 import { existsSync } from 'node:fs';
 import { readdir } from 'node:fs/promises';
+import { homedir } from "os";
```
B4nan, line 9:
```suggestion
import { homedir } from 'os';
```

test/e2e/tools.mjs
```diff
@@ -30,6 +31,12 @@ export async function getStats(url) {
     return fs.readJSON(path);
 }
 
+export async function getApifyToken() {
+    const authPath = join(homedir(), '.apify', 'auth.json');
+    const { token } = await fs.readJSON(authPath);
```
B4nan, line 36: we might want to check if the file exists first (and throw a hint that login is required to run e2e tests), this would probably fail if not

B4nan (approved):
few nits before we merge

test/e2e/tools.mjs
AndreyBykov: oh jeez, I got too used to eslint =D

B4nan:
image

PR #26, opened 2022-01-31
chore: split all crawlers into their respective packages

vladfrangu:
For now, this is marked as draft while I try to debug why turborepo breaks hard from this PR...


packages/core/src/main.ts
```diff
@@ -45,7 +42,7 @@ export {
 
     AutoscaledPool,
 
-    BasicCrawler,
+    // BasicCrawler,
```
B4nan, line 45: i see - this should definitely go, we should not export things from here that we are moving to their own packages, all the crawler related exports should go, and we should import them from the basic crawler package instead (and as said on slack, going forward we should split that and change imports to the leaf packages).

vladfrangu: Honestly.. the exports in core are a mess, we should really clean them up completely. Is this file or exports.ts even needed anymore?

packages/puppeteer-crawler/src/internals/stealth/index.ts
```diff
@@ -1,10 +1,12 @@
...
-import globalLog from '../utils_log';
+import {
+    logUtils,
+    keys,
+} from '@crawlers/core';
 import * as hidingTricks from './hiding_tricks';
-import { keys } from '../typedefs';
 
-const log = globalLog.child({ prefix: 'Stealth' });
+const log = logUtils.child({ prefix: 'Puppeteer Stealth' });
```
B4nan, line 9: lets keep the prefix
```suggestion
const log = logUtils.child({ prefix: 'Stealth' });
```

B4nan (changes requested):

packages/core/src/crawlers/index.ts
```diff
@@ -1,8 +1,8 @@
...
-export * from './cheerio_crawler';
+export * from './crawler_commons';
+// export * from './browser_crawler';
+// export * from './cheerio_crawler';
 export * from './crawler_extension';
 export * from './crawler_utils';
-export * from './playwright_crawler';
-export * from './puppeteer_crawler';
+// export * from './playwright_crawler';
+// export * from './puppeteer_crawler';
```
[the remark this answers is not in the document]
vladfrangu: Check the path, its the core module's crawler file, not the crawlers package :D
However, those comments shouldn't have been there to begin with, so out they'll go

packages/puppeteer-crawler/src/internals/stealth/index.ts
vladfrangu: I mean right now it only applies to puppeteer, and if you use both puppeteer and playwright crawlers..you could think stealth applies to both. I will remove it, but just making sure

packages/core/src/crawlers/index.ts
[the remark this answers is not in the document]
B4nan: yeah you are right, lets remove those lines

packages/puppeteer-crawler/src/internals/stealth/index.ts
B4nan: but this will never log anything for playwright, or will it? its in the puppeteer package

vladfrangu: It won't, you're right, but shouldn't we implement it for Playwright too?

B4nan: the plan is actually to remove it in v3 in favour or fingerprints

vladfrangu: Oh. Well then 😁
I'll apply the change then 👍

packages/browser-crawler/src/internals/browser-launcher.ts
```diff
@@ -1,29 +1,18 @@
...
-
-    /**
-     * If `true` and `executablePath` is not set,
-     * Playwright will launch full Google Chrome browser available on the machine
-     * rather than the bundled Chromium. The path to Chrome executable
-     * is taken from the `APIFY_CHROME_EXECUTABLE_PATH` environment variable if provided,
-     * or defaults to the typical Google Chrome executable location specific for the operating system.
-     * By default, this option is `false`.
-     * @default false
-     */
```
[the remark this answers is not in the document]
vladfrangu: I think I forgot to add it back woops. Will re-add

packages/playwright-crawler/package.json
```diff
@@ -35,15 +40,30 @@
...
+    },
+    "peerDependencies": {
+        "playwright": "^1.11.0"
+    },
+    "peerDependenciesMeta": {
+        "playwright": {
+            "optional": true
+        }
+    },
+    "devDependencies": {
```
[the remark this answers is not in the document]
vladfrangu: I might be able to remove the devDependency stuff actually, it seemed like npm figured out what went where, so will look into it

packages/puppeteer-crawler/src/internals/puppeteer-crawler.ts
```diff
@@ -223,8 +229,8 @@ export class PuppeteerCrawler extends BrowserCrawler<LaunchOptions, { browserPlu
 
         if (stealth) {
             this.browserPool.postLaunchHooks.push(async (_pageId, browserController) => {
-                // @TODO: We can do this better now. It is not necessary to override the page.
-                //   we can modify the page in the postPageCreateHook
+                // TODO: We can do this better now. It is not necessary to override the page.
+                // we can modify the page in the postPageCreateHook
```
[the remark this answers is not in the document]
vladfrangu: VSC doesn't have such features, it just renders the comments as is, and extra whitespaces just for webstorm feel off.. I can revert this if you want 👍

packages/browser-crawler/src/internals/browser-launcher.ts
```diff
@@ -1,29 +1,27 @@
...
      * or defaults to the typical Google Chrome executable location specific for the operating system.
      * By default, this option is `false`.
      * @default false
      */
     useChrome?: boolean;
-
-    /**
-     * By default this function uses `require("playwright").chromium`.
-     * If you want to use a different browser you can pass it by this property as `require("playwright").firefox`
-     */
```
B4nan, line 26: same here, we should adjust the comment if its dated, not remove it

vladfrangu: It doesn't really make sense to have these on the base interface... Puppeteer doesn't have a browserType iirc, playwright does

vladfrangu: Oh, I see, for puppeteer we have it typed as unknown due to puppeteer-extra and such. Thoughts?

packages/puppeteer-crawler/src/internals/puppeteer-crawler.ts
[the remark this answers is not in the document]
B4nan: well, what i said above does not feel off at all, i also see this as two separate comments, lets keep it please. 

btw "just for webstorm" feels like gatekeeping, not everyone likes vscode 😜 it actually adopts features that webstorm/idea pioneered ages ago. and they are still not all there.

image
image

packages/browser-crawler/src/internals/browser-launcher.ts
B4nan: > It doesn't really make sense to have these on the base interface... Puppeteer doesn't have a browserType iirc, playwright does

Then let's move it to the place where it makes sense

packages/playwright-crawler/package.json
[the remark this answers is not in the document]
B4nan: what about this one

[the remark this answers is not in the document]
vladfrangu: Seems like it isn't needed but we need to be careful with semver ranges once we start publishing them and updating after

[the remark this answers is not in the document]
B4nan: well, if it was wrong before and you needed this hack to make the build work... it is a good thing we know it was wrong before, the build should fail in such case.

B4nan (approved):
lets merge. just one hint for future - we dont need _any_ dev deps in child packages, only those in the root package.json make sense, we should add all of them there

PR #35, opened 2022-02-15
feat: context aware enqueueLink

### LINKED ISSUE #28: Move `enqueueLinks` to the crawling context
Currently to use `enqueueLinks` we have the global helper and we need to pass it all the information ourselves. The goal is to have the `enqueueLinks` method accessible on the crawling context and be context aware, so we don't have to pass things like RequestQueue instance to it. 

Old API:

```js
const requestQueue = await Apify.openRequestQueue();
await requestQueue.addRequest({ url: 'https://www.iana.org/' });

const crawler = new Apify.PuppeteerCrawler({
    requestQueue,
    handlePageFunction: async ({ request, page }) => {
        const title = await page.title();
        console.log(`Title of ${request.url}: ${title}`);
        await Apify.utils.enqueueLinks({
            requestQueue,
            page,
            pseudoUrls: ['https://www.iana.org/[.*]'],
        });
    },
});
```

Proposed API:

```js
const requestQueue = await Apify.openRequestQueue();
await requestQueue.addRequest({ url: 'https://www.iana.org/' });

const crawler = new Apify.PuppeteerCrawler({
    requestQueue,
    handlePageFunction: async ({ request, page, enqueueLinks }) => {
        const title = await page.title();
        console.log(`Title of ${request.url}: ${title}`);
        await enqueueLinks({
            pseudoUrls: ['https://www.iana.org/[.*]'],
        });
    },
});
```

vladfrangu:
Closes #28

### DIFF
--- a/packages/core/src/enqueue_links/enqueue_links.ts
+++ b/packages/core/src/enqueue_links/enqueue_links.ts
@@ -1,6 +1,4 @@
-import ow, { ArgumentError } from 'ow';
-import { URL } from 'url';
-import log from '../utils_log';
+import ow from 'ow';
 import {
     constructPseudoUrlInstances,
     createRequests,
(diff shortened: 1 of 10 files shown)


packages/core/src/enqueue_links/enqueue_links.ts
```diff
@@ -143,50 +119,27 @@ export async function enqueueLinks(options: EnqueueLinksOptions): Promise<QueueO
...
-    const pseudoUrlInstances = constructPseudoUrlInstances(pseudoUrls as PseudoUrlInput[] || []);
+    let parsedPseudoUrls: PseudoUrlInput[];
+
+    if (Array.isArray(pseudoUrls)) {
+        parsedPseudoUrls = pseudoUrls;
+    } else if (pseudoUrls) {
+        parsedPseudoUrls = [pseudoUrls];
+    } else {
+        parsedPseudoUrls = [];
+    }
```
B4nan, line 131: why is this needed? the ow validation seems to allow only null and an array, so the middle branch seems not reachable

packages/core/src/enqueue_links/enqueue_links.ts
```diff
@@ -10,24 +8,14 @@ import {
     PseudoUrlInput,
 } from './shared';
 import { RequestQueue, QueueOperationInfo } from '../storages/request_queue';
-import { PseudoUrl } from '../pseudo_url';
 import { validators } from '../validators';
-import { CheerioRoot } from '../utils';
 
-export interface EnqueueLinksOptions {
+export interface BaseEnqueueLinksOptions {
```
B4nan, line 13: let's keep the interface name here as it was
```suggestion
export interface EnqueueLinksOptions {
```

B4nan (approved):
almost perfact, just few nits

PR #39, opened 2022-02-16
chore: migrate storage-local to monorepo

AndreyBykov:
So I think it's pretty much done. I had to update two request queue tests in `storage-local` as they failed with a weird error: 

> Expected constructor: ArgumentError
> Received constructor: ArgumentError

So instead of `expect(err).toBeInstanceOf(ArgumentError)` I do `expect(err.toString()).toMatch('ArgumentError')` now - not sure if we could leave it this way for now.

Otherwise, all tests pass on the platform and locally (although I still had to restart them as random windows ones still fail every now and then).


B4nan:
> So instead of expect(err).toBeInstanceOf(ArgumentError) I do expect(err.toString()).toMatch('ArgumentError') now - not sure if we could leave it this way for now.

This is almost certainly because you haven't upgraded the dependencies, and rest of the packages use newer version of `ow`. You end up with multiple installs, so multiple `ArgumentError` classes. They are not compatible even if they have the same code - as they are two distinct files.

test/storage-local/datasets.test.ts
```diff
@@ -0,0 +1,343 @@
+import { emptyDirSync, ensureDirSync, readdirSync, readFile, writeJsonSync } from 'fs-extra';
+import { join } from 'path';
+import { ApifyStorageLocal } from '@apify/storage-local';
+import { STORAGE_NAMES } from '@apify/storage-local/src/consts';
```
B4nan, line 4: dont use deep imports

```suggestion
import { ApifyStorageLocal, STORAGE_NAMES } from '@apify/storage-local';
```

if the symbol is not exported, export it

test/storage-local/index.test.ts
```diff
@@ -0,0 +1,78 @@
+import log from '@apify/log';
+import { join } from 'path';
+import { ensureDirSync, statSync, writeFileSync } from 'fs-extra';
+import { ApifyStorageLocal } from '@apify/storage-local';
+import { STORAGE_NAMES } from '@apify/storage-local/src/consts';
```
B4nan, line 5: same here, no deep imports

test/storage-local/request_queues.test.ts
```diff
@@ -0,0 +1,854 @@
+import { ensureDirSync, readdirSync } from 'fs-extra';
+import { ArgumentError } from 'ow';
+import { join } from 'path';
+import type { Database, Statement } from 'better-sqlite3-with-prebuilds';
+import { ApifyStorageLocal } from '@apify/storage-local';
+import { STORAGE_NAMES, DATABASE_FILE_NAME } from '@apify/storage-local/src/consts';
+import { RequestQueueEmulator } from '@apify/storage-local/src/emulators/request_queue_emulator';
+import { uniqueKeyToRequestId } from '@apify/storage-local/src/utils';
+import type { DatabaseConnectionCache } from '@apify/storage-local/src/database_connection_cache';
+import type { RequestModel } from '@apify/storage-local/src/resource_clients/request_queue';
```
B4nan, line 10: and again

test/storage-local/utils.test.ts
```diff
@@ -0,0 +1,37 @@
+import * as utils from '@apify/storage-local/src/utils';
```
B4nan, line 1: here as well

B4nan (changes requested):
i see this package will need some love, the code is quite dirty, but let's not care about it right now :]

AndreyBykov:
After discussion on Slack - updated in accordance with the comments except for the deep imports for now.

PR #42, opened 2022-02-21
refactor: make the concept of storage clients abstract

B4nan:
WIP

### DIFF
--- a/packages/apify/src/actor.ts
+++ b/packages/apify/src/actor.ts
@@ -3,7 +3,6 @@ import { ENV_VARS, INTEGER_ENV_VARS } from '@apify/consts';
 import log from '@apify/log';
 import { ActorRun as ClientActorRun, ActorStartOptions, ApifyClient, ApifyClientOptions, TaskStartOptions, Webhook, WebhookEventType } from 'apify-client';
 import {
-    ActorRunWithOutput,
     Configuration,
     ConfigurationOptions,
     Dataset,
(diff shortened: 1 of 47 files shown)


packages/apify/src/actor.ts
```diff
@@ -1352,6 +1351,40 @@ export interface MetamorphOptions {
...
+     * ```
+     * {
+     *   "contentType": "application/json; charset=utf-8",
+     *   "body": {
+     *     "message": "Hello world!"
+     *   }
+     * }
+     * ```
+     */
+    output?: Dictionary | null;
```
mnmkng, line 1387: I would keep the `call` API exactly the same as Client now. The final version of how this whole output management will look like is still undecided. So there should be no need for this type.

B4nan: This PR is not about that, the diff is just about moving types from a different package to the actor sdk.

But good catch, this is indeed forgotten type that is wrong (there is no `fetchOutput` anymore in the code), will remove it.

mnmkng: Not sure I understand. You can't have docs for a method you don't have a separate interface for?

B4nan: Sure we can, but the client type has no jsdoc at all for `ActorRun`, and it does not make much sense to have jsdoc talking about the `Actor` class from SDK in client.

mnmkng: I think having the return value documented in the method will be enough.

PR #43, opened 2022-02-22
feat: utils package

vladfrangu:


packages/utils/src/index.ts
```diff
@@ -0,0 +1,8 @@
+export * from './internals/cheerio';
```
B4nan, line 1: can you describe this convention with `internals` folder? it seems like everything is an internal, yet we export everything, so... are they really internals? :]

i know there is one approach with two barrel files, one for public exports (index.ts) and one for all exports (internal.ts), which might be a way to get around import cycles, but never really used internals as a folder

packages/utils/src/internals/cheerio.ts
```diff
@@ -0,0 +1,86 @@
+import cheerio, { load } from 'cheerio';
```
B4nan, line 1: i would move this to the cheerio package, wdyt? given how small the crawler packages are, we can just export those directly, or have a namespace export like `export * as utils`

packages/utils/src/internals/extract-urls.ts
```diff
@@ -0,0 +1,72 @@
+import ow from 'ow';
```
B4nan, line 1: `ow` is not declared as dependency

packages/utils/src/internals/memory-info.ts
```diff
@@ -0,0 +1,166 @@
+// @ts-expect-error We need to add typings for @apify/ps-tree
+import psTree from '@apify/ps-tree';
```
B4nan, line 2: `@apify/ps-tree` is undeclared dependency

packages/utils/src/internals/social.ts
```diff
@@ -568,30 +549,3 @@ export const YOUTUBE_REGEX = new RegExp(`^${YOUTUBE_REGEX_STRING}$`, 'i');
  * ```
  */
 export const YOUTUBE_REGEX_GLOBAL = new RegExp(YOUTUBE_REGEX_STRING, 'ig');
-
-/** @internal */
-export const socialUtils = {
```
B4nan, line 573: we might still use a namespace export to group all the regexes, will confirm

test/apify/crawlers/browser_crawler.test.ts
```diff
@@ -867,7 +881,7 @@ describe('BrowserCrawler', () => {
             await browserCrawler.run();
 
             delete process.env[ENV_VARS.PROXY_PASSWORD];
-            stub.restore();
+            stub.mockClear();
```
B4nan, line 884: fyi we generally need to restore, not just clear the mocks. FYI there is `jest.restoreAllMocks()` we can run in beforeEach hook and just forgot about this (could be even part of some global test setup).

test/apify/crawlers/browser_crawler.test.ts
```diff
@@ -22,11 +22,25 @@ import {
     RequestList,
     sleep,
     createProxyConfiguration,
-    requestUtils,
+    requestAsBrowser,
 } from 'crawlers';
 import LocalStorageDirEmulator from '../local_storage_dir_emulator';
 import { BrowserCrawlerTest } from './basic_browser_crawler';
 
+jest.mock('@crawlers/utils/src/internals/request', () => {
```
B4nan, line 30: uh, ugly, but we will get rid of this when we use got scraping directly i hope. never really liked the partial mocking, always preferred `jest.spyOn` :]

B4nan (changes requested):
good job!

jest.config.js
```diff
@@ -20,6 +20,7 @@ module.exports = {
         '^@crawlers/cheerio$': '<rootDir>/packages/cheerio-crawler/src',
         '^@crawlers/playwright$': '<rootDir>/packages/playwright-crawler/src',
         '^@crawlers/puppeteer$': '<rootDir>/packages/puppeteer-crawler/src',
+        '^@crawlers/utils$': '<rootDir>/packages/utils/src',
```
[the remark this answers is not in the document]
vladfrangu: Ah, alright, I thought we just wanted to always define them strictly when we added a new package

packages/apify/src/index.ts
```diff
@@ -1 +1,2 @@
 export * from './actor';
+export * from './utils';
```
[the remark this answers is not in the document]
vladfrangu: Makes sense, will update

packages/apify/src/utils.ts
```diff
@@ -0,0 +1,42 @@
+import { log } from '@crawlers/core';
+import { ENV_VARS } from '@apify/consts';
+import { type } from 'node:os';
```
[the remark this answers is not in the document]
vladfrangu: 😂 didn't even see that haha, if you want me to not deconstruct it lmk

packages/core/src/utils.ts
```diff
@@ -469,7 +14,7 @@ export function htmlToText(html: string | CheerioRoot): string {
  */
 export function createRequestDebugInfo(
     request: Request,
-    response: IncomingMessage | Partial<PuppeteerResponse> = {},
+    response: IncomingMessage | Partial<BrowserResponseLike> = {},
```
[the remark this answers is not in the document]
vladfrangu: I guess I could move it in the request.ts file instead

packages/core/src/autoscaling/snapshotter.ts
```diff
@@ -3,9 +3,9 @@ import ow from 'ow';
 import { betterSetInterval, betterClearInterval, BetterIntervalID } from '@apify/utilities';
 import { ACTOR_EVENT_NAMES, ENV_VARS } from '@apify/consts';
 import { Log } from '@apify/log';
-import { getMemoryInfo, isAtHome } from '../utils';
+import { log as defaultLog, getMemoryInfo } from '@crawlers/utils';
+// import { isAtHome } from '../utils';
```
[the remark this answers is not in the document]
vladfrangu: There already is a skipped test for it, so I suppose we can remove it

packages/utils/src/index.ts
vladfrangu: Well, in most places/modules I do, I use `src/lib` for the actual code, but `lib` is gitignored here, so I chose the next thing. We can change that, of course, if you'd like

packages/utils/src/internals/cheerio.ts
vladfrangu: Doesn't really make sense, as its not a cheerio crawler specific util... I guess its iffy

test/apify/crawlers/browser_crawler.test.ts
vladfrangu: I actually tried using jest.spyOn, but because the import differences, it didn't apply :c (and I spent way too long debugging it only to have a 💡 moment)

B4nan: yeah i know that feeling, sometimes its more sane to add a new method just for easier mocking :] 

let's see when we use got scraping directly

packages/apify/src/utils.ts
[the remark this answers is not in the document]
B4nan: its fine like this, it just made me laugh so wanted to share :]

packages/core/src/utils.ts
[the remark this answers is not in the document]
B4nan: looks like its used only in scraper tools? if so, i'd move it there

packages/utils/src/index.ts
B4nan: we can improve on that later, i would maybe just put the files in src directly. btw that gitignored lib folder sounds like some historical thing, we can definitely make it not ignored, i dont see why not

packages/utils/src/internals/cheerio.ts
B4nan: how come its not cheerio specific when it uses cheerio? i see it is used in some other utils, but those also look cheerio specific and should be moved to the cheerio package i'd say

packages/core/src/utils_log.ts
```diff
@@ -1,73 +0,0 @@
...
- * You can customize your logging further by extending or replacing the default
- * logger instances with your own implementations.
- *
- * @module log
- */
-
-import log, { Log, LoggerOptions, LogLevel, Logger, LoggerJson, LoggerText } from '@apify/log';
-
-// TODO move this to @apify/log? check if we still need this, probably yes
-Object.assign(log, { Log, LogLevel, Logger, LoggerJson, LoggerText });
```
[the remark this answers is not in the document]
vladfrangu: So I should remove the utils re-export?

packages/utils/src/internals/cheerio.ts
vladfrangu: Well, you can pass in any strings to it (so you could technically use a puppeteer page, get the html and pass it to this util)... but yeah I can see the confusion it can cause.. its your call really

B4nan: will get some feedback from Ondra, for me the goal is to not depend on cheerio anywhere except the cheerio crawler (or have it as optional peer dependency, but that feels more like a hack for this case)

(definitely not a blocker for merging this PR)

packages/crawlers/src/index.ts
```diff
@@ -1,4 +1,34 @@
 export * from '@crawlers/core';
+// This export doesn't use * because of type collisions with /core, yet re-exporting manually works
```
B4nan, line 2: i would just `@ts-ignore` this if we want to reexport everything, this sounds error prone in such case

but what are the conflicts? maybe we should not export something from the core package? we dont really need to export things like `keys`, `entries`, `Constructor` or `Dictionary`, those are purely internal (and if something, they can be exported from one of those packages, not from both)

vladfrangu: That's a good point, I'll fix the core exports instead 👍

packages/core/src/autoscaling/snapshotter.ts
```diff
@@ -195,13 +195,8 @@ export class Snapshotter {
         // Start snapshotting.
         this.eventLoopInterval = betterSetInterval(this._snapshotEventLoop.bind(this), this.eventLoopSnapshotIntervalMillis);
         this.clientInterval = betterSetInterval(this._snapshotClient.bind(this), this.clientSnapshotIntervalMillis);
-        if (isAtHome()) {
-            events.on(ACTOR_EVENT_NAMES.SYSTEM_INFO, this._snapshotCpuOnPlatform);
-            events.on(ACTOR_EVENT_NAMES.SYSTEM_INFO, this._snapshotMemoryOnPlatform);
-        } else {
-            this.cpuInterval = betterSetInterval(this._snapshotCpuOnLocal.bind(this), this.cpuSnapshotIntervalMillis);
-            this.memoryInterval = betterSetInterval(this._snapshotMemoryOnLocal.bind(this), this.memorySnapshotIntervalMillis);
-        }
```
B4nan, line 204: let's not change this, we need to abstract the platform events, not just ignore them :]

packages/core/src/autoscaling/snapshotter.ts
```diff
@@ -472,13 +467,9 @@ export class Snapshotter {
...
         if (this.maxMemoryBytes) return;
         const { totalBytes } = await this._getMemoryInfo();
-        if (isAtHome()) {
-            this.maxMemoryBytes = totalBytes;
-        } else {
-            this.maxMemoryBytes = Math.ceil(totalBytes / 4);
-            // NOTE: Log as AutoscaledPool, so that users are not confused what "Snapshotter" is
-            this.log.info(`Setting max memory of this run to ${Math.round(this.maxMemoryBytes / 1024 / 1024)} MB. Use the ${ENV_VARS.MEMORY_MBYTES} environment variable to override it.`); // eslint-disable-line max-len
-        }
+        this.maxMemoryBytes = Math.ceil(totalBytes / 4);
```
B4nan, line 470: same here, i will try to deal with the `isAtHome` thingy in the other PR, but if we remove it, it will be easy to forget about it (and will break the platform usage)

B4nan (changes requested):
where are all those import reordering changes coming from? it makes the review 10 times longer than it needs to be, and i can already see the merge conflicts :]

packages/core/src/autoscaling/snapshotter.ts
vladfrangu: Will bring them back but commented out like before, is that alright?

B4nan: Sure, just add a `FIXME` comment there (I do prefer those for things that really should not get published, while `TODO` is something we can improve on later).

vladfrangu:
These windows tests are gonna kill me 🥲

B4nan (approved):

B4nan:
But next time please dont use any autoformatting tools, if we want that, it needs to be integrated to the repository, so we are all aligned. 

image

vladfrangu:
Ouch. I can do the rebase if you want, but I didn't expect that many conflicts 😓

B4nan:
I'll deal with it this time. Note that order of imports can have effect on what we compile and can result in cycles that show up only after compilation to CJS. Reordering of imports can be actually dangerous and harmful, I've been there :]

Issue #32, opened 2022-02-14
Improve error handling for request retries


B4nan:
Currently every failed request ends up with a huge and scary error in the log, even if we plan on retrying the request. 

We should only print a one line warning about this, and print an error only if we no longer plan on retrying. We should distinguish known errors here and also remove stack trace from things like timeouts, as the stack trace has no added value. For now let's do timeout use case, next step should be to do known http status code errors, e.g. 403, 404.

This should be new default behaviour, but we want to allow printing the full errors with stack traces via some configuration. Let's make this work based on some env var (let's call it just `VERBOSE_RETRY_LOGS`, we can improve on that later if needed).

B4nan:
More errors where we dont care about stack trace:

```
2022-02-14T12:48:24.605Z   RequestError: The proxy server rejected the request with status code 502
2022-02-14T12:48:24.611Z       at ClientRequest.<anonymous> (/usr/src/app/node_modules/got-cjs/dist/source/core/index.js:820:126)
2022-02-14T12:48:24.618Z       at Object.onceWrapper (node:events:510:26)
2022-02-14T12:48:24.623Z       at ClientRequest.emit (node:events:402:35)
2022-02-14T12:48:24.631Z       at ClientRequest.origin.emit (/usr/src/app/node_modules/@szmarczak/http-timer/dist/source/index.js:43:20)
2022-02-14T12:48:24.644Z       at ClientRequest.request.emit (/usr/src/app/node_modules/got-scraping/dist/hooks/fix-decompress.js:76:20)
2022-02-14T12:48:24.652Z       at emitErrorNT (node:internal/streams/destroy:157:8)
2022-02-14T12:48:24.656Z       at emitErrorCloseNT (node:internal/streams/destroy:122:3)
2022-02-14T12:48:24.661Z       at processTicksAndRejections (node:internal/process/task_queues:83:21)
2022-02-14T12:48:24.665Z       at Http2OverHttp.createConnection (/usr/src/app/node_modules/http2-wrapper/source/proxies/h2-over-hx.js:19:10)
2022-02-14T12:48:24.669Z       at runMicrotasks (<anonymous>)
2022-02-14T12:48:24.673Z       at processTicksAndRejections (node:internal/process/task_queues:96:5)
2022-02-14T12:48:24.678Z       at async entry (/usr/src/app/node_modules/http2-wrapper/source/agent.js:399:15)
```

```
2022-02-14T12:47:19.660Z   Error: Got status code 404
[...]

PR #44, opened 2022-02-22
feat: improve error handling for request retries

AndreyBykov:
By default:
- for timeout errors:
  - log warning WITHOUT stack trace for requests, that would be retried;
  - log error WITHOUT stack trace after the last retry;
- for non-timeout errors:
  - log warning WITHOUT stack trace for requests, that would be retried;
  - log error WITH stack trace after the last retry;

With log level set to DEBUG:
- keep the old behavior, i.e. log error with stack trace for requests that would be retried and after last failed retry.


packages/basic-crawler/src/internals/basic-crawler.ts
```diff
@@ -697,11 +697,21 @@ export class BasicCrawler<
...
             const { url, retryCount, id } = request;
-            this.log.exception(
-                error,
-                'handleRequestFunction failed, reclaiming failed request back to the list or queue',
-                { url, retryCount, id },
-            );
+            // We don't want to see the stack trace in the logs by default, when we are going to retry the request.
+            // Thus, we print the full stack trace only for DEBUG log level.
+            // Maybe it's worth adding another ENV variable instead of using log levels here.
+            if (this.log.getLevel() < 5) {
```
vladfrangu, line 703: Can you use the enum level instead of the magic number? 🙏

vladfrangu (commented):
Looks fine to me, main nitpick is this

AndreyBykov:
I also guess we should keep https://github.com/apify/apify-ts/issues/32 open as this PR only covers the timeout error, not status codes?

B4nan:
Yeah ideally handle it as part of the PR or a follow up one before we close it. I would do 40x and 50x errors, I guess we can have one error type for 40x and one for 50x, no need for 10+ classes (but the error instance needs to hold the http error code). 

Not sure if we currently retry all those, we should definitely make this configurable, so users can pick how to threat what error type, so the solution should be generic for all the error types.

AndreyBykov:
Ok, cool - so let's keep this one open - I will continue with statuses.

packages/basic-crawler/src/internals/basic-crawler.ts
```diff
@@ -724,9 +726,11 @@ export class BasicCrawler<
             await this.failedContextHandler(crawlingContext);
         } else {
             const { id, url, method, uniqueKey } = crawlingContext.request;
-            this.log.exception(
-                crawlingContext.error,
-                'Request failed and reached maximum retries',
+            this.log.error(
+                `Request failed and reached maximum retries. ${crawlingContext.error instanceof TimeoutError && this.log.getLevel() < 5
```
szmarczak, line 730: Can we use a constant variable instead of `5`?

szmarczak (approved):

test/apify/crawlers/basic_crawler.test.ts
```diff
@@ -641,6 +664,198 @@ describe('BasicCrawler', () => {
...
+            handleRequestFunction: () => sleep(10),
+        });
+
+        const warnMock = jest.fn();
+        // @ts-expect-error Overriding protected method
+        crawler.log.warning = warnMock;
+
+        const errMock = jest.fn();
+        // @ts-expect-error Overriding protected method
+        crawler.log.error = errMock;
```
vladfrangu, line 684: Can't you use jest.spyOn instead? `const warningSpy = jest.spyOn(crawler.log, 'warning')` etc

packages/basic-crawler/src/internals/basic-crawler.ts
vladfrangu, line 730:
```suggestion
                `Request failed and reached maximum retries. ${crawlingContext.error instanceof TimeoutError && this.log.getLevel() < this.log.LEVELS.DEBUG
```

vladfrangu (changes requested):
Just one small nit as well as Szymon's comment and a rebase

packages/basic-crawler/src/internals/basic-crawler.ts
AndreyBykov: I was pretty sure I fixed this one before, probably messed something up when I was making more changes. Fixed.

test/apify/crawlers/basic_crawler.test.ts
AndreyBykov: it's pretty much the same, but yeah - I guess it's more readable/clean. Updated.

vladfrangu (approved):

B4nan:
@AndreyBykov please rebase this one, I'd like to merge it. Maybe it's not exactly the solution we wanted to have, but it's definitely an improvement in the right direction.

AndreyBykov:
Yup, we even discussed that we need to figure out how to move forward. So yeah - it's definitely more like a first step. Will do 👍

packages/basic-crawler/src/internals/basic-crawler.ts
```diff
@@ -696,10 +696,12 @@ export class BasicCrawler<
...
             const { url, retryCount, id } = request;
-            this.log.exception(
-                error,
-                'handleRequestFunction failed, reclaiming failed request back to the list or queue',
-                { url, retryCount, id },
+            // We don't want to see the stack trace in the logs by default, when we are going to retry the request.
+            // Thus, we print the full stack trace only for DEBUG log level.
+            // Maybe it's worth adding another ENV variable instead of using log levels here.
+            this.log.warning(
+                `Reclaiming failed request back to the list or queue. ${this.log.getLevel() < this.log.LEVELS.DEBUG ? error : error.stack}`,
```
B4nan, line 703: let's use env var for this instead of log levels, maybe `CRAWLEE_VERBOSE_LOG` or something along that

packages/basic-crawler/src/internals/basic-crawler.ts
```diff
@@ -723,9 +725,11 @@ export class BasicCrawler<
             await this.failedContextHandler(crawlingContext);
         } else {
             const { id, url, method, uniqueKey } = crawlingContext.request;
-            this.log.exception(
-                crawlingContext.error,
-                'Request failed and reached maximum retries',
+            this.log.error(
+                `Request failed and reached maximum retries. ${crawlingContext.error instanceof TimeoutError && this.log.getLevel() < this.log.LEVELS.DEBUG
```
B4nan, line 729: same here

B4nan (changes requested):
just one thing, i'd rather not use the log levels

AndreyBykov:
@B4nan done.

AndreyBykov:
I've renamed `handleRequestFunction` to `requestHandler` in errors in basic crawler and some other places - did not want to leave it for some other PR, as it's quite misleading to keep it with old prop name.

AndreyBykov:
Otherwise done.

test/core/crawlers/basic_crawler.test.ts
```diff
@@ -577,6 +600,174 @@ describe.each(SingleStorageCase)('BasicCrawler - %s', (Emulator) => {
...
+            expect(/Request failed and reached maximum retries/.test(args[0])).toBe(true);
+            expect(/Other non-timeout error/.test(args[0])).toBe(true);
+            expect(/at BasicCrawler\.requestHandler/.test(args[0])).toBe(true);
+            expect(args[1]).toBeDefined();
+        }
+    });
+
+    test('should log stack trace for timeout errors when verbose log is enabled', async () => {
+        log.setLevel(log.LEVELS.INFO);
+        process.env.CRAWLEE_VERBOSE_LOG = String(true);
```
B4nan, line 684: lol
```suggestion
        process.env.CRAWLEE_VERBOSE_LOG = 'true';
```

test/core/crawlers/basic_crawler.test.ts
```diff
@@ -577,6 +600,174 @@ describe.each(SingleStorageCase)('BasicCrawler - %s', (Emulator) => {
...
+            expect(args[1]).toBeDefined();
+        }
+
+        log.setLevel(log.LEVELS.OFF);
+        process.env.CRAWLEE_VERBOSE_LOG = undefined;
+    });
+
+    test('should log stack trace for non-timeout errors when verbose log is enabled', async () => {
+        log.setLevel(log.LEVELS.INFO);
+        process.env.CRAWLEE_VERBOSE_LOG = String(true);
```
B4nan, line 728:
```suggestion
        process.env.CRAWLEE_VERBOSE_LOG = 'true';
```

test/core/crawlers/basic_crawler.test.ts
AndreyBykov: Now I remember I did it once already 😄 it was late - my head was not working properly 🙃 jeez 😄

PR #47, opened 2022-03-02
feat: rename handleRequest/handlePage* to requestHandler*

### LINKED ISSUE #33: Rename `handlePageFunction` to `requestHandler`
Instead of `handlePageFunction` and `handleRequestFunction` we wil lhave `requestHandler`, instead of `handleFailedRequestFunction` there will be `failedRequestHandler`. 

We should keep the old options there to throw an exception saying that the method name changed. Should be marked as `@deprecated` in the interfaces. 

This applies to all related `handlePage*` options and class properties, e.g. `handlePageTimeoutSecs`.

vladfrangu:
Closes #33


packages/basic-crawler/src/internals/basic-crawler.ts
```diff
@@ -270,10 +328,9 @@ export class BasicCrawler<
     autoscaledPool?: AutoscaledPool;
 
     protected log: Log;
-    protected userProvidedHandler: HandleRequest<Context>;
-    protected failedContextHandler?: HandleFailedRequest<ErrorContext>;
-    protected handleFailedRequestFunction?: HandleFailedRequest<ErrorContext>;
-    protected handleRequestTimeoutMillis: number;
+    protected userProvidedHandler!: RequestHandler<Context>;
```
B4nan, line 331: lets rename this, the handler is always "user provided", i'd use just `requestHandler`. also let's inline `_handleRequestHandler`.

btw i thought this will be much easier, so many changes :P

vladfrangu: Inlining it will be fun, since the method gets overridden in classes that extend it. Should be doable

B4nan: if it makes sense to keep it, lets keep it, i just saw it here without any purpose. but the name is quite meh, "handle handler" :]

vladfrangu: I'd say let's keep it (since extensions of basic crawler can augment the context), and I agree, the naming could be improved.. runRequestHandler maybe?

packages/actor-scraper/cheerio-scraper/src/internals/crawler_setup.ts
```diff
@@ -159,7 +159,7 @@ export class CrawlerSetup implements CrawlerSetupOptions {
 
         const options: CheerioCrawlerOptions = {
             proxyConfiguration: this.proxyConfiguration,
-            handlePageFunction: this._handlePageFunction.bind(this),
+            requestHandler: this._handlePageFunction.bind(this),
```
B4nan, line 162: `_handlePageFunction` should be renamed too

B4nan, line 162: `_handlePageFunction` should be renamed too

packages/actor-scraper/cheerio-scraper/src/internals/crawler_setup.ts
```diff
@@ -168,7 +168,7 @@ export class CrawlerSetup implements CrawlerSetupOptions {
             prepareRequestFunction: this._prepareRequestFunction.bind(this),
             requestTimeoutSecs: this.input.pageLoadTimeoutSecs,
             ignoreSslErrors: this.input.ignoreSslErrors,
-            handleFailedRequestFunction: this._handleFailedRequestFunction.bind(this),
+            failedRequestHandler: this._handleFailedRequestFunction.bind(this),
```
B4nan, line 171: `_handleFailedRequestFunction` -> `_failedRequestHandler`

B4nan, line 171: `_handleFailedRequestFunction` -> `_failedRequestHandler`

packages/basic-crawler/src/internals/basic-crawler.ts
```diff
@@ -764,6 +868,44 @@ export class BasicCrawler<
             await this.sessionPool!.teardown();
         }
     }
+
+    protected _handlePropertyNameChange<New, Old>({
```
B4nan, line 872: this looks quite crazy :] maybe it would be easier to have multiple (but much simpler) methods, one for each deprecated property? that way you dont need that `CrawlerRenames` thingy at all

B4nan, line 872: this looks quite crazy :] maybe it would be easier to have multiple (but much simpler) methods, one for each deprecated property? that way you dont need that `CrawlerRenames` thingy at all

B4nan (changes requested):

packages/basic-crawler/src/internals/basic-crawler.ts
B4nan: I don't like to have those `CrawlerRenames` props, so that is what I would like to get rid of mostly.

vladfrangu: I've removed the `CrawlerRenames` in favor of an implementation similar to the one done in BrowserCrawler

PR #51, opened 2022-03-09
chore: test update (sinon + underscore)

barjin:
`sinon` and `underscore` were removed from the tests.

It seems that Sinon is more capable (and less verbose) than Jest, but the base idea behind the tests should stay the same.


vladfrangu (approved):

test/apify/request_list.test.ts
```diff
@@ -619,7 +630,7 @@ describe('RequestList', () => {
             reqs.push(request);
         }
 
-        reqs = shuffle(reqs);
+        reqs = shuffle(reqs) as typeof reqs;
```
B4nan, line 633: nit: if you add a generic type argument to the function, you won't need to cast anything

e.g. something like this:

```ts
shuffle<T extends unknown[]>(a: T): T { ... }
```

B4nan (approved):

PR #52, opened 2022-03-09
chore: update examples for enqueueLinks

vladfrangu:
image
image


website/src/css/customTheme.css
```diff
@@ -139,3 +139,14 @@ aside svg[class*=iconExternalLink] {
 header.hero div[class^=heroButtons] {
     justify-content: inherit;
 }
+
+.docusaurus-highlight-code-line {
+	background-color: rgba(0, 0, 0, 0.1);
```
B4nan, line 144: we will need editorconfig to enforce the whitespaces. i dont mind tabs, i do mind mixed usage

(i was actually long time tab advocate, used it for a decade before moving to JS where nobody understood why tabs are great, which is finally changing these days)

PR #60, opened 2022-03-16
feat: support batched add requests in `RequestQueue `

vladfrangu:
I've only done preliminary tests, however it's mostly a copy-paste of addRequest with added handling for multiple requests

Closes https://github.com/apify/apify-ts/issues/58

### DIFF
--- a/test/storage-local/request_queues.test.ts
+++ b/test/storage-local/request_queues.test.ts
@@ -88,7 +88,7 @@ describe('sanity checks for seeded data', () => {
     });
 
     test('seeded queues exist', () => {
-        expect(counter.queues()).toBe(2);
+        expect(counter.queues()).toBe(3);
     });
 
     Object.values(TEST_QUEUES).forEach((queue) => {
@@ -331,7 +331,7 @@ describe('getOrCreate', () => {
         const queue = await storageLocal.requestQueues().getOrCreate('first');
         expect(queue.id).toBe('first');
         const count = counter.queues();
-        expect(count).toBe(2);
+        expect(count).toBe(3);
     });
 
     test('creates a new queue with name', async () => {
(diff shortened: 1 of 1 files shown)


vladfrangu:
Ah, Actions have some issues... That explains why CI is not running rn
image

B4nan (changes requested):
i would like to see some integration tests, seems like you only tested the client method is called, but not that it actually works

@valekjo could you please take a peek at this? also how does the `unprocessedRequests` work, do we need to handle that somehow here? do you think we should make this opt in and fallback to something like `Promise.add(reqs.map(req => rq.addRequest(req))` (in batches)? this is v3, so it wont be stable at least for a month or two, and we will be doing quite a lot of testing before it gets stable, so we will catch issues early during internal testing in case it would be broken on api end

vladfrangu:
> i would like to see some integration tests, seems like you only tested the client method is called, but not that it actually works

I'd love to setup an E2E test for this, maybe I can get some help from @AndreyBykov with doing it if he has time? Would also let us see if the endpoint works

B4nan:
Mmmm actually, I am quite sure it will not work, as those tests work locally, and we don't have the batch support implemented in storage-local! Please take a look at that too, do it in the https://github.com/apify/apify-storage-local-js repo, not in the monorepo copy, so we can publish it.

vladfrangu:
> also how does the `unprocessedRequests` work, do we need to handle that somehow here?

At a first glance, it seems to return requests that weren't added to the request queue (at a glance it seems to be requests that already exist in the queue that would be filled in here). I don't think we want to re-add them again forcefully

test/storage-local/request_queues.test.ts
```diff
@@ -508,6 +514,72 @@ describe('addRequest', () => {
     });
 });
 
+test.only('addRequests', async () => {
```
B4nan, line 517:
```suggestion
test('addRequests', async () => {
```

vladfrangu: Woops, thx

PR #62, opened 2022-03-21
fix: remove requestAsBrowser in favor of got-scraping

### LINKED ISSUE #34: use got-scraping directly instead of requestAsBrowser
- remove compatibility layer in `requestAsBrowser` 
- add `got` export to `got-scraping` and use that in the docs
- adjust all the examples and jsdocs
- write down small migration guide about the differences between `requestAsBrowser` and using got-scraping directly

szmarczak:
Fixes #34

- [x] remove compatibility layer in `requestAsBrowser` 
- [ ] ~remove `got` export in `got-scraping` or replace it with `gotScraping`~ (out of scope of this PR)
- [x] adjust all the examples and jsdocs
- [ ] write down small migration guide about the differences between `requestAsBrowser` and using got-scraping directly


szmarczak:
The code is still quite messy. Needs a refactoring (should be a separate PR) - the `context.response` is expected to be `IncomingMessage`, however that isn't right with Got ideology. Got wraps parts of `ClientRequest` and `IncomingMessage` into Got's `Request`. For example, in order to access the original `IncomingMessage`, you need `request.response`. So, in order to get the headers, `request.response.headers`.

To solve this, there has been added [`addResponsePropertiesToStream`](https://github.com/apify/apify-ts/blob/c3850ecdb8024cd9e7e80dbcde97a47a39121968/packages/utils/src/internals/request.ts#L427) which turns Got's `Request` instance to look like `IncomingMessage`, overwriting some properties along the way. Oh, and it gets casted from `Request` to `IncomingMessage`.

docs/examples/basic_crawler.mdx
```diff
@@ -10,8 +10,8 @@ import BasicCrawlerSource from '!!raw-loader!./basic_crawler.ts';
 This is the most bare-bones example of the Apify SDK, which demonstrates some of its building blocks such as the <ApiLink to="basic-crawler/class/BasicCrawler">`BasicCrawler`</ApiLink>. You probably don't need to go this deep though, and it would be better to start with one of the full-featured crawlers
 like <ApiLink to="cheerio-crawler/class/CheerioCrawler">`CheerioCrawler`</ApiLink> or <ApiLink to="playwright-crawler/class/PlaywrightCrawler">`PlaywrightCrawler`</ApiLink>.
 
-The script simply downloads several web pages with plain HTTP requests using the <ApiLink to="utils/function/requestAsBrowser">`requestAsBrowser()`</ApiLink>
-convenience function and stores their raw HTML and URL in the default dataset. In local configuration, the data will be stored as JSON files in
+The script simply downloads several web pages with plain HTTP requests using the `got-scraping`
```
B4nan, line 13: let's use a link to the got-scraping repository

docs/examples/crawl_single_url.mdx
```diff
@@ -7,8 +7,8 @@ import CodeBlock from '@theme/CodeBlock';
 import ApiLink from '@site/src/components/ApiLink';
 import CrawlSource from '!!raw-loader!./crawl_single_url.ts';
 
-This example uses the <ApiLink to="utils/function/requestAsBrowser">`requestAsBrowser()`</ApiLink>
-function to grab the HTML of a web page.
+This example uses the `got-scraping` npm package
```
B4nan, line 10: let's use a link to the got-scraping repository

packages/basic-crawler/src/internals/basic-crawler.ts
```diff
@@ -277,7 +279,13 @@ export interface BasicCrawlerOptions<
  *     handleRequestFunction: async ({ request }) => {
  *         // 'request' contains an instance of the Request class
  *         // Here we simply fetch the HTML of the page and store it to a dataset
- *         const { body } = await Actor.utils.requestAsBrowser(request);
+ *         const { body } = await gotScraping({
+ *             url: request.url,
+ *             method: request.method,
+ *             body: request.payload,
```
B4nan, line 285: we need to do something about this, so we support passing request instances directly. either at the SDK level, or in got-scraping. i guess it would be simpler to support `body` as an alias to payload in got-scraping, right?

packages/cheerio-crawler/src/internals/cheerio-crawler.ts
```diff
@@ -850,26 +847,24 @@ export class CheerioCrawler<JSONData = unknown> extends BasicCrawler<
...
 
         try {
-            responseWithStream = await this._requestAsBrowser(opts);
+            return await this._requestAsBrowser(opts);
         } catch (e) {
             if (e instanceof TimeoutError) {
                 this._handleRequestTimeout(session);
-            } else {
-                throw e;
+                return undefined as unknown as IncomingMessage;
```
B4nan, line 856: ouch :D looking at the `_handleRequestTimeout` implementation, i would vote to inline it so TS understands we throw in this branch.

packages/cheerio-crawler/src/internals/cheerio-crawler.ts
```diff
@@ -1089,7 +1091,7 @@ interface RequestFunctionOptions {
     request: Request;
     session?: Session;
     proxyUrl?: string;
-    requestAsBrowserOptions: RequestAsBrowserOptions;
+    requestAsBrowserOptions: OptionsInit;
```
B4nan, line 1094: the property should be renamed to match what its doing

B4nan (approved):

PR #63, opened 2022-03-22
feat: update enqueueLinks

### LINKED ISSUE #45: Remove `PseudoUrl` in favour of glob patterns
- move `PseudoUrl` implementation to new package in apify-shared-js
  - add also `purlToRegExp(purl: string): RegExp` function that will convert string pseudo url to regexp instance
- use this package in `enqueueLinks` (via `purlToRegExp` function), mark its usage as deprecated and suggets using `globs` instead
- allow using glob patterns (via [minimatch](https://github.com/isaacs/minimatch)) and regexes directly

```ts
enqueueLinks({
    globs: [
        'https://example.com/**',
        'https://example.com/?*/?*',
    ],
});

// or

enqueueLinks({
    regexps: [
        /https:\/\/example\.com.*/,
        /https:\/\/example\.com\/.+\/.+/,
    ],
});

// or

enqueueLinks({
    pseudoUrls: [
        'https://example.com/*',
        'https://example.com/*/*',
    ],
}),
```

In future (v4) we will remove the `pseudoUrls` usage and allow only explict `purlToRegExp` via `regexps`

AndreyBykov:
Still, WIP, having some problems with tests + need to update tests in `shared` + need to merge previous changes.

Also - thinking to update `actors-scrapers`, as in the end I think it's rather a mess to leave `requestOptions` binded to regexp. I first wanted to put it everywhere (to globs/regexps/pseudoUrls), but `transformRequestFunction` seems like a better solution.

So please have a look, while I would continue with the tests.

Closes #45


packages/browser-crawler/src/internals/browser-crawler.ts
```diff
@@ -593,9 +594,10 @@ interface EnqueueLinksInternalOptions {
     page: CommonPage;
     requestQueue?: RequestQueue;
     defaultBaseUrl?: string;
+    transformRequestFunction?: RequestTransform;
 }
 
-export async function browserCrawlerEnqueueLinks({ options, page, requestQueue, defaultBaseUrl }: EnqueueLinksInternalOptions) {
+export async function browserCrawlerEnqueueLinks({ options, page, requestQueue, defaultBaseUrl, transformRequestFunction }: EnqueueLinksInternalOptions) {
```
vladfrangu, line 600: Any reason for this? You can already pass it down in options, right?

packages/core/src/enqueue_links/enqueue_links.ts
```diff
@@ -123,43 +138,74 @@ export async function enqueueLinks(options: EnqueueLinksOptions): Promise<QueueO
...
+    }
 
     if (options.baseUrl) {
         switch (options.strategy ?? EnqueueStrategy.SameDomain) {
             case EnqueueStrategy.SameDomain:
                 // We need to get the origin of the passed in domain in the event someone sets baseUrl
                 // to a url like https://example.com/deep/default/path and one of the found urls is an
                 // absolute relative path (/path/to/page)
-                extraPseudoUrls.push(`${new URL(options.baseUrl).origin}/[.*]`);
+                regexps.push(new RegExp(new URL(options.baseUrl).origin));
```
vladfrangu, line 194: This won't match the same although it might be fine? or maybe we should specifically make it match the start of the string?

test/apify/enqueue_links/enqueue_links.test.ts
```diff
@@ -150,14 +161,22 @@ describe('enqueueLinks()', () => {
...
                 page,
                 requestQueue,
+                transformRequestFunction: (request) => {
+                    if (/example\.com/.test(request.url)) {
+                        request.method = 'POST';
+                    } else if (/cool\.com/.test(request.url)) {
+                        request.userData.foo = 'bar';
+                    }
+                    return request;
+                },
```
B4nan, line 179: btw i would still keep a test like this too, so we know it works both ways

test/apify/enqueue_links/enqueue_links.test.ts
```diff
@@ -223,25 +242,16 @@ describe('enqueueLinks()', () => {
...
+            try {
+                await browserCrawlerEnqueueLinks({
+                    // @ts-expect-error Type 'RegExp[]' is not assignable to type 'PseudoUrlInput[]'
+                    options: { selector: '.click', pseudoUrls },
+                    page,
+                    requestQueue,
+                });
+            } catch (err) {
+                expect((err as Error).message).toMatch(/Expected values to be of type `string` but received type `RegExp`/);
+            }
```
B4nan, line 254: for testing error handling, you should prefer `expect(() => ....).toThrow()` as that way you know things will actually throw or the test will fail. the way you have it now would silently pass if the function would not throw.

test/apify/enqueue_links/enqueue_links.test.ts
```diff
@@ -356,9 +366,9 @@ describe('enqueueLinks()', () => {
             };
 
             const pseudoUrls = [
-                new PseudoUrl('https://example.com/[(\\w|-|/)*]', { method: 'POST' }),
+                new PseudoUrl('https://example.com/[(\\w|-|/)*]'),
                 null,
-                new PseudoUrl('[http|https]://cool.com/', { userData: { foo: 'bar' } }),
+                new PseudoUrl('[http|https]://cool.com/'),
```
B4nan, line 371: same here. 

i would also expect the `null` to actually trigger validation, i guess that was some kind of hack in case parsing failed, but ignoring that seems like a wrong thing to do

packages/core/src/enqueue_links/enqueue_links.ts
B4nan: yeah this needs to be definitly fixed, you can use `purlToRegExp` to build the regexp and inline it here

packages/core/src/enqueue_links/shared.ts
```diff
@@ -1,5 +1,5 @@
 import { URL } from 'url';
-import { PseudoUrl, PseudoUrlObject } from '../pseudo_url';
+import { PseudoUrl } from '@apify/pseudo_url';
```
B4nan, line 2:
```suggestion
import { purlToRegExp } from '@apify/pseudo_url';
```

packages/core/src/enqueue_links/shared.ts
```diff
@@ -13,57 +13,57 @@ const MAX_ENQUEUE_LINKS_CACHE_SIZE = 1000;
  */
 const enqueueLinksPseudoUrlCache = new Map();
 
-export type PseudoUrlInput = string | RegExp | PseudoUrl | PseudoUrlObject;
+export type PseudoUrlObject = { purl: string };
+
+export type PseudoUrlInput = string | PseudoUrl | PseudoUrlObject;
```
B4nan, line 18:
```suggestion
export type PseudoUrlInput = string | PseudoUrlObject;
```

test/apify/enqueue_links/enqueue_links.test.ts
```diff
@@ -115,12 +115,23 @@ describe('enqueueLinks()', () => {
             };
 
             const pseudoUrls = [
-                new PseudoUrl('https://example.com/[(\\w|-|/)*]', { method: 'POST' }),
-                new PseudoUrl('[http|https]://cool.com/', { userData: { foo: 'bar' } }),
+                new PseudoUrl('https://example.com/[(\\w|-|/)*]'),
+                new PseudoUrl('[http|https]://cool.com/'),
```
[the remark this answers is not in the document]
AndreyBykov: Yeah, that's what I was kinda hesitating the most - I went this way first, but in my head it kinda was a bit messy code-wise and when I started to think about it - I though that well - let's keep the `transformRequestFunction`, but ok, will update.

test/apify/enqueue_links/enqueue_links.test.ts
AndreyBykov: yup, will do both

test/apify/enqueue_links/enqueue_links.test.ts
AndreyBykov: makes sense, will update, thx

packages/browser-crawler/src/internals/browser-crawler.ts
AndreyBykov: you're right, I was kinda trying to quickly fix the test and missed it. Will update (but it would probably change with other changes, not sure yet). Thanks for noticing!

test/apify/enqueue_links/enqueue_links.test.ts
AndreyBykov: Not sure what you mean here. If there's just `null` (i.e. no patterns provided) - it will mean that all links found on page will be enqueued. If there are some patterns in an array and one of them is `null` - it kinda goes against each other (enqueue matching vs enqueue all) - so it seems reasonable to me that it should throw. Or you meant something else?

Otherwise ready for review.

szmarczak (commented):
LGTM code-wise.

packages/core/src/enqueue_links/enqueue_links.ts
```diff
@@ -148,15 +185,38 @@ export async function enqueueLinks(options: EnqueueLinksOptions): Promise<QueueO
...
+    }
 
     if (options.baseUrl) {
         switch (options.strategy ?? EnqueueStrategy.SameSubdomain) {
             case EnqueueStrategy.SameSubdomain:
                 // We need to get the origin of the passed in domain in the event someone sets baseUrl
                 // to a url like https://example.com/deep/default/path and one of the found urls is an
                 // absolute relative path (/path/to/page)
-                extraPseudoUrls.push(`${new URL(options.baseUrl).origin}/[.*]`);
+                urlPatternObjects.push({ regexp: purlToRegExp(`${new URL(options.baseUrl).origin}/[.*]`) });
```
vladfrangu, line 219: Can't we convert this to native regexes / globs now, instead of using purl? (or should we handle that in a future PR)

packages/core/src/enqueue_links/shared.ts
```diff
@@ -1,68 +1,149 @@
...
+    return regexps.map((item) => {
+        // Get regexp object from cache.
+        let regexpObject = enqueueLinksPatternCache.get(item);
+        if (regexpObject) return regexpObject;
+
+        if (item instanceof RegExp) {
+            regexpObject = { regexp: item };
+        } else {
+            const { regexp, ...requestOptions } = item;
+            regexpObject = { regexp, ...requestOptions };
```
vladfrangu, line 114: Isn't this a bit redundant? Or did you mean to create a RegExp from the deconstructed one?

vladfrangu (changes requested):
Two small nits, LGTM otherwise

packages/core/src/enqueue_links/enqueue_links.ts
AndreyBykov: Updated :+1:

packages/core/src/enqueue_links/shared.ts
AndreyBykov: It is redundant - went from globs from above and just used the same pattern. So you're right, it's not necessary here.

packages/core/src/enqueue_links/enqueue_links.ts
B4nan: This is not correct, as you are not escaping special characters in the origin, and it will definitely contain things like dots, question marks or plus signs.

AndreyBykov: Or do you mean `purlToRegExp` is not correct? But I haven't changed the logic - i.e. the hole `purlToRegExp` is the same as it was before...

AndreyBykov: Went for `${new URL(options.baseUrl).origin}/**` glob and not for `${new URL(options.baseUrl).origin}/**/*` because first option would allow to not have anything after `[domain]/` while second with `**/*` would expect something here -> not the same behaviour as before.

Also for `EnqueueStrategy.SameHostname` `${url.origin.replace(baseUrlHostname, *.${baseUrlHostname})}/**` with `**/*` in the end would not allow urls with different prefix, but without deeper levels. E.g. for `www.apify.com` it would not allow something like `sdk.apify.com` or `help.apify.com` and would expect something after `.com`.

test/apify/enqueue_links/enqueue_links.test.ts
B4nan: I dont understand why we should even allow the `null` as a valid input, thats what I mean. Imho `null` should trigger validation = be disallowed. I dont see a single reason why it would be good idea to keep it, given we now have enqueing strategies that are more explicit. 

And the "same here" was about using `PseudoUrl` constructors, that's what i said above is not necessary and should be removed from supported input too - currently there is no benefit in this, the class does not do anything else than buiding the regexp and testing it.

On the other hand the test was originally using custom `userData` and the one before was having `POST` method, both of which should be still supported - but no longer via the `PseudoUrl` class. we should still test this.

B4nan (changes requested):
once again for clarity:
- `null` should _not_ be accepted input for `pseudoUrls`
- `PseudoUrl` class should _not_ be used anywhere (not even in tests), only `purlToRegExp`

test/apify/enqueue_links/enqueue_links.test.ts
AndreyBykov: Adjusted the `null` option :+1:

Rest (`PseudoUrl` constructor) was already removed from everywhere.

AndreyBykov: And yeah - tests were also adjusted.

test/apify/enqueue_links/enqueue_links.test.ts
```diff
@@ -141,56 +144,73 @@ describe('enqueueLinks()', () => {
             expect(enqueued[2].userData.foo).toBe('bar');
         });
 
-        test('works with Actor UI output object', async () => {
```
B4nan, line 144: oh so this is where it comes from :D

test/apify/enqueue_links/enqueue_links.test.ts
```diff
@@ -288,29 +338,11 @@ describe('enqueueLinks()', () => {
                 enqueued.push(request);
             };
 
-            await browserCrawlerEnqueueLinks({
+            await expect(async () => browserCrawlerEnqueueLinks({
```
B4nan, line 341: you can do just this when testing async methods with `expect().resolves/rejects`
```suggestion
            await expect(browserCrawlerEnqueueLinks({
```

B4nan (changes requested):
_few_ final comments, we are getting there :]

packages/core/src/enqueue_links/shared.ts
```diff
@@ -79,7 +159,8 @@ export function createRequestOptions(sources: (string | Record<string, unknown>)
             } catch (err) {
                 return false;
             }
-        });
+        })
+        .map((requestOptions) => new Request(requestOptions) as RequestOptions);
```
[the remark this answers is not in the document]
AndreyBykov: @B4nan the reason for this is that request creates `userData` and proper request properties, while creating an object does not do it. So I thought instead of creating `userData` explicitly I would just create a request. On the other hand - you are right that there are requestOptions method and request creation method. So I guess let's create the `userData` explicitly for requestOptions - will update it

packages/core/src/enqueue_links/shared.ts
```diff
@@ -1,68 +1,148 @@
...
+    if (globTrimmed.length === 0) throw new Error(`Cannot parse Glob pattern '${globTrimmed}': it must be an non-empty string`);
+    return globTrimmed;
+}
+
+/**
+ * Helper factory used in the `enqueueLinks()` and enqueueLinksByClickingElements() function
+ * to check RegExps input and return valid RegExps.
+ * @ignore
+ */
+export function constructRegExpObjectsFromRegExps(regexps: RegExpInput[]): RegExpObject[] {
```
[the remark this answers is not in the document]
AndreyBykov: But it's used e.g. in [enqueueLinksByClickingElements](https://github.com/apify/apify-ts/blob/master/packages/puppeteer-crawler/src/internals/enqueue-links/click-elements.ts), where it's imported from `@crawlers/core`. For now there's `constructRegExpObjectsFromPseudoUrls` only, but I guess we also need to add globs/regexps here (and other related stuff - disallow null patterns etc). And if the `construct...ObjectsFrom...` functions will not be exported from `core` - should we just basically copy/paste them here (it partially repeats the `enqueueLinks` logic...)? That does not seem right :/

test/apify/enqueue_links/enqueue_links.test.ts
```diff
@@ -106,7 +106,7 @@ describe('enqueueLinks()', () => {
             expect(enqueued[3]).toBe(undefined);
         });
 
-        test('works with PseudoUrl instances', async () => {
+        test('works with Actor UI output object', async () => {
```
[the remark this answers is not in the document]
AndreyBykov: I was actually always confused here - I thought it's rather `Actor UI input object` - still not sure why it's called UI output... But basically it about testing whether it works with pattern objects... On the other hand - it's covered in `works with...` tests - so maybe completetly remove these?

packages/core/src/enqueue_links/shared.ts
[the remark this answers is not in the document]
AndreyBykov: Updated, but it still feels a bit weird with a check for userData existence and creating an empty object if it's not there.

packages/core/src/enqueue_links/enqueue_links.ts
```diff
@@ -31,17 +36,48 @@ export interface EnqueueLinksOptions {
...
      *
-     * The plain objects must include at least the `purl` property, which holds the pseudo-URL string or RegExp.
-     * All remaining keys will be used as the `requestTemplate` argument of the {@link PseudoUrl} constructor,
-     * which lets you specify special properties for the enqueued {@link Request} objects.
+     * The matching is always case-insensitive.
+     * If you need case-sensitive matching, use `regexps` property directly.
      *
-     * If `pseudoUrls` is an empty array, `null` or `undefined`, then the function
+     * If `globs` is an empty array or `undefined`, then the function
      * enqueues all links found on the page.
```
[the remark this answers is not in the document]
AndreyBykov: Other jsdoc - are you referring to `click-elements.ts`? If yes - please check the comment below - and yeah - I should also update it (but it's the same stuff from `enqueueLinks`...

packages/core/src/enqueue_links/shared.ts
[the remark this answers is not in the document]
B4nan: ok lets keep it as it is right now, i thought its just for the tests. 

i know this wasn't introduced in this PR, rather in the strategies support refactor... but it feels wrong/weird to have so many publicly exported helper methods - those are just implementation details. maybe we could export all of them under some namespace.

packages/core/src/enqueue_links/shared.ts
```diff
@@ -79,6 +161,10 @@ export function createRequestOptions(sources: (string | Record<string, unknown>)
             } catch (err) {
                 return false;
             }
+        })
+        .map((requestOptions) => {
+            if (!requestOptions.userData) requestOptions.userData = {};
```
B4nan, line 166: we can do better now :]

```suggestion
            requestOptions.userData ??= {};
```

B4nan (approved):
let's merge, we can clean up the tests later

B4nan:
so one last thing - rebase please

PR #67, opened 2022-04-01
feat: Actor.reboot()

### LINKED ISSUE #65: Add `Actor.reboot()` to restart the container on the Apify platform
Sometimes actors running on the Apify platform get into a state that they cannot easily recover from, like a memory leak. This can easily be fixed by aborting and resurrecting the actor, but that needs to be done manually. By adding `Actor.reboot()` we could enable this functionality from code.

The function should:
1. pause any running crawlers (it needs to accept those in options)
2. persist state (optionally, default true)
3. metamorph the actor into itself; this will restart the container while keeping the same `runId`, storages etc.

barjin:
closes #65

### DIFF
--- a/packages/apify/src/actor.ts
+++ b/packages/apify/src/actor.ts
@@ -283,10 +283,9 @@ export class Actor {
      * Internally reboots this actor. The system stops the current container and starts
      * a new container with the same run ID.
      *
-     * @param [_options]
      * @ignore
      */
-    async reboot(_options: RebootOptions = {}): Promise<void> {
+    async reboot(): Promise<void> {
         if (!this.isAtHome()) {
             log.warning('Actor.reboot() is only supported when running on the Apify platform.');
             return;
@@ -302,8 +301,8 @@ export class Actor {
             ],
         );
 
-        const actId = this.config.get('actorId');
-        await this.metamorph(actId!);
+        const actorId = this.config.get('actorId');
+        await this.metamorph(actorId!);
     }
 
     /**
@@ -888,8 +887,8 @@ export class Actor {
      *
      * @param [options]
      */
-    static async reboot(options: RebootOptions = {}): Promise<void> {
-        return Actor.getDefaultInstance().reboot(options);
+    static async reboot(): Promise<void> {
+        return Actor.getDefaultInstance().reboot();
     }
 
     /**
@@ -1366,10 +1365,6 @@ export interface MetamorphOptions {
 
     /** @internal */
     customAfterSleepMillis?: number;
-}
-
-export interface RebootOptions {
-
 }
 
 export interface ExitOptions {
--- a/packages/apify/src/actor.ts
+++ b/packages/apify/src/actor.ts
@@ -3,10 +3,12 @@ import { ENV_VARS, INTEGER_ENV_VARS } from '@apify/consts';
 import log from '@apify/log';
 import { ActorRun as ClientActorRun, ActorStartOptions, ApifyClient, ApifyClientOptions, TaskStartOptions, Webhook, WebhookEventType } from 'apify-client';
 import {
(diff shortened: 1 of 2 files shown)


packages/apify/src/actor.ts
```diff
@@ -268,13 +270,42 @@ export class Actor {
...
+            return;
+        }
+
+        // Waiting for all the listeners to finish, as `.metamorph()` kills the container.
+        await Promise.all(
+            [
+                // `persistState` for individual RequestLists, RequestQueue... instances to be persisted
+                ...events.listeners(ACTOR_EVENT_NAMES_EX.PERSIST_STATE).map((x) => x()),
+                // `migrating` to pause Apify crawlers
+                ...events.listeners(ACTOR_EVENT_NAMES_EX.MIGRATING).map((x) => x()),
```
B4nan, line 301: @mnmkng do we need to emit both events here? i believe internally migrating event is just a flavor of persist_state, so the migrating event could be enough? 

https://github.com/apify/apify-ts/blob/master/packages/core/src/events.ts#L104

i was also expecting just `events.emit()` call, the metamorph call waits for 5 minutes before it resolves, which i believe was the reason - to wait for the events to finish. i am not against explicit await like this, but i guess it would be better done in the `metamorph` method directly, and there might still be some other reason for that long sleep inside

B4nan (changes requested):

packages/apify/src/actor.ts
barjin: IIRC `ACTOR_EVENT_NAMES.MIGRATING` is caught only in [`BasicCrawler`](https://github.com/apify/apify-ts/blob/c3850ecdb8024cd9e7e80dbcde97a47a39121968/packages/basic-crawler/src/internals/basic-crawler.ts#L521). 
Don't know what the best practices are, but can imagine a user using only the `RequestList` etc. separately within their own solution - `RequestList` is listening only to the `PERSIST_STATE` event.

mnmkng: @B4nan Dunno. I would double check against your new event implementation :)

The long sleep is there so that the container does not abruptly exit while the API performs the metamoph.

B4nan: I guess we need both, as they are fired manually.

barjin: Also, the sleep call happens only after the `metamorph` API call, when the metamorph is already happening.

I'm just guessing here, but if metamorph call means that the Apify platform performs `docker stop` on our container -> an IPC signal to the main process (`SIGTERM`, grace period, then `SIGKILL`). `SIGTERM`, if uncaught, terminates the node process immediately without waiting for any `async` functions. This is still an uneducated guess, but this sounds like a potential data loss/corruption in case of a long "saving" process and overly eager platform.

packages/apify/src/actor.ts
```diff
@@ -268,13 +270,42 @@ export class Actor {
...
+            ],
+        );
+
+        const actId = this.config.get('actorId');
```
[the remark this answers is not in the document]
barjin: noted :) 📝

test/apify/actor.test.ts
```diff
@@ -394,6 +394,33 @@ describe('Actor.metamorph()', () => {
...
+    });
+
+    afterEach(() => {
+        delete process.env[ENV_VARS.IS_AT_HOME];
+        delete process.env[ENV_VARS.ACTOR_ID];
+        delete process.env[ENV_VARS.ACTOR_RUN_ID];
+        jest.restoreAllMocks();
+    });
+
+    test('metamorphs to the same actor', async () => {
```
B4nan, line 413: would be good to test that the wait for async events works

B4nan (approved):

PR #68, opened 2022-04-04
chore: migrate @apify/browser-pool to monorepo

AndreyBykov:


B4nan:
Looking like the recent changes in browser pool are not yet backported.

Namely this one https://github.com/apify/browser-pool/pull/70

packages/browser-pool/package.json
```diff
@@ -22,7 +22,7 @@
 	"license": "Apache-2.0",
 	"repository": {
 		"type": "git",
-		"url": "git+https://github.com/apify/browser-pool-js"
+        "url": "https://github.com/apify/browser-pool.git"
```
B4nan, line 25: this should reflect the new repository, also version should be `3.0.0` as other packages

AndreyBykov:
Yep, I knew about the changes. All there now. Also the comment from 3 days ago about CI - seems like the workflows are not using the master, but are using the current branch. Repository/version number updated.

There also were 2 missing awaits - not sure if it was done intentionally, added them rather then just adding ts-ignore -> [this](https://github.com/apify/apify-ts/pull/68/commits/f7a4be277848da6e01fd5a64e973a11b0ec14eec) commit.

packages/browser-pool/src/browser-pool.ts
```diff
@@ -749,9 +749,9 @@ export class BrowserPool<
         if (browserController.activePages === 0 && this.retiredBrowserControllers.has(browserController)) {
             // Run this with a delay, otherwise page.close()
             // might fail with "Protocol error (Target.closeTarget): Target closed."
-            setTimeout(() => {
+            setTimeout(async () => {
```
B4nan, line 752: i dont think this is a good idea, you stil trigger the async callback without awaiting it. lets add a TODO here to revisit this

AndreyBykov: Reverted to the previous state. Added TODO + eslint-disable-line

AndreyBykov:
@B4nan all green :ok_hand:

packages/browser-pool/package.json
```diff
@@ -0,0 +1,53 @@
...
+		"fingerprint-generator": "^1.0.0",
+		"fingerprint-injector": "^1.0.0",
+		"lodash.merge": "^4.6.2",
+		"nanoid": "^3.3.1",
+		"ow": "^0.28.0",
+		"p-limit": "^3.1.0",
+		"quick-lru": "^5.1.1",
+		"proxy-chain": "^2.0.1-beta.0",
+		"tiny-typed-emitter": "^2.1.0",
+		"tslib": "^2.3.1"
```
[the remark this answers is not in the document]
AndreyBykov: There are still quite a few versions used. `core` is using 1.1.6, and some other packages are using 1.1.10. I would use the current latest, or should I stick to 1.1.6?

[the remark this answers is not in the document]
B4nan: yop, lets use latest stable

PR #70, opened 2022-04-06
feat: update enqueueLinksByClickingElements

AndreyBykov:
Follow-up to #63 
- add `globs`/`regexps` options for `enqueueLinksByClickingElements`, deprecate `pseudoUrls` (same as `enqueueLinks`)
- small update to `enqueueLinks` jsdoc
- remove `Actor UI output object` tests for `enqueueLinks` (they were about input objects, it's now covered by other tests).

### DIFF
--- a/packages/core/src/enqueue_links/enqueue_links.ts
+++ b/packages/core/src/enqueue_links/enqueue_links.ts
@@ -38,8 +38,9 @@ export interface EnqueueLinksOptions {
     /**
      * An array of glob pattern strings or plain objects
      * containing glob pattern strings matching the URLs to be enqueued.
-     * All remaining keys will be used as request options
-     * for the corresponding enqueued {@link Request} objects.
+     *
+     * The plain objects must include at least the `glob` property, which holds the glob pattern string.
+     * All remaining keys will be used as request options for the corresponding enqueued {@link Request} objects.
      *
      * The matching is always case-insensitive.
      * If you need case-sensitive matching, use `regexps` property directly.
@@ -52,8 +53,9 @@ export interface EnqueueLinksOptions {
     /**
      * An array of regular expressions or plain objects
      * containing regular expressions matching the URLs to be enqueued.
-     * All remaining keys will be used as request options
-     * for the corresponding enqueued {@link Request} objects.
+     *
+     * The plain objects must include at least the `regexp` property, which holds the glob pattern string.
+     * All remaining keys will be used as request options for the corresponding enqueued {@link Request} objects.
      *
      * If `regexps` is an empty array or `undefined`, then the function
      * enqueues the links with the same subdomain.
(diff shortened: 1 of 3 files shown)


packages/core/src/enqueue_links/enqueue_links.ts
```diff
@@ -66,8 +68,9 @@ export interface EnqueueLinksOptions {
      *
      * An array of {@link PseudoUrl} strings or plain objects
```
B4nan, line 69: We no longer use this class, so it won't be present in the docs and the link most probably won't work. Maybe we should move the content from the [JSDoc of that class](https://github.com/apify/apify-shared-js/blob/master/packages/pseudo_url/src/pseudo_url.ts#L4) to some other place so it gets rendered in our docs - we dont need the full content as is, but we should keep the part that describes what PURL is and how it works somewhere (could be a separate docs page, doesn't have to be in code/jsdoc).

But I guess we don't have to care

packages/core/src/enqueue_links/enqueue_links.ts
```diff
@@ -66,8 +68,9 @@ export interface EnqueueLinksOptions {
...
      * containing {@link PseudoUrl} strings matching the URLs to be enqueued.
-     * All remaining keys will be used as request options
-     * for the corresponding enqueued {@link Request} objects.
+     *
+     * The plain objects must include at least the `purl` property, which holds the glob pattern string.
```
B4nan, line 72: `purl` property does not hold glob pattern

packages/core/src/enqueue_links/enqueue_links.ts
```diff
@@ -52,8 +53,9 @@ export interface EnqueueLinksOptions {
     /**
      * An array of regular expressions or plain objects
      * containing regular expressions matching the URLs to be enqueued.
-     * All remaining keys will be used as request options
-     * for the corresponding enqueued {@link Request} objects.
+     *
+     * The plain objects must include at least the `regexp` property, which holds the glob pattern string.
```
B4nan, line 57: `regexp` property does not hold glob pattern

packages/puppeteer-crawler/src/internals/enqueue-links/click-elements.ts
```diff
@@ -49,15 +54,52 @@ export interface EnqueueLinksByClickingElementsOptions {
...
+    globs?: GlobInput[];
+
+    /**
+     * An array of regular expressions or plain objects
+     * containing regular expressions matching the URLs to be enqueued.
+     *
+     * The plain objects must include at least the `regexp` property, which holds the glob pattern string.
+     * All remaining keys will be used as request options for the corresponding enqueued {@link Request} objects.
+     *
+     * If `regexps` is an empty array or `undefined`, then the function
```
B4nan, line 79: we will need to revisit those docs carefully, as it works differently now (again, by default, same subdomain strategy is used)

i will add a task to the roadmap, no need to deal with this here/now

AndreyBykov: There's no strategy at all. It works with clicking elements, and there's no default `a` selector. So it works with all links produced by page after clicking the elements with provided selector. It's either then checks it vs patterns or enqueues it all.

So it's not just copied from enqeueLinks:

> If `pseudoUrls` is an empty array or `undefined`, then the function
> enqueues all the intercepted navigation requests produced by the page
> after clicking on elements matching the provided CSS selector.

packages/core/src/enqueue_links/enqueue_links.ts
B4nan: nope, it wont be part of the monorepo

AndreyBykov: Welp :] Then let's go with your plan above =D

Issue #64, opened 2022-03-23
Fix `enqueueLinks` with explicit pseudourls and redirecting urls


B4nan:
Currently the enqueing strategy is always used, we should ignore that if some explicit pseudourls/regexps/globs are provided in the options. 

We also need to rework how the defaultBaseUrl works, we need to have both `request.url` and `request.loadedUrl` in the options, `request.url` being used for enqueing (it is before redirects, which we want for the enqueing base), and `request.loadedUrl` used for the relative links in cheerio crawler.

Lastly, we should not have default for the request queueu in the arguments, it should be always provided either by user or from the context, no default and we should validate that it is there via `ow` in the lowest level implementation.

PR #72, opened 2022-04-10
refactor: make `Actor.main()` awaitable and rework one E2E test as an example

B4nan:
- `BasicCrawler.run()` now returns final statistics
- `Actor.main()` now returns promise, so it can be awaited (still throws synchronously for validation)
- `Actor.main()` is now generic and is typed to return the same as the `UserFunc` callback returns
- ~new method `Actor.run` that calls `crawler.run()` inside `Actor.main` for easier testing~
- fixes `enqueueLinks` default strategy being applied even with explicit pseudoUrls
- fixes creating crawlers with implicit `RequestQueue`

### DIFF
--- a/packages/basic-crawler/src/internals/basic-crawler.ts
+++ b/packages/basic-crawler/src/internals/basic-crawler.ts
@@ -24,6 +24,7 @@ import {
     Configuration,
     EventManager,
     EventType,
+    FinalStatistics,
 } from '@crawlers/core';
 import { Awaitable } from '@crawlers/utils';
 import ow, { ArgumentError } from 'ow';
@@ -338,7 +339,6 @@ export class BasicCrawler<
     protected sessionPoolOptions: SessionPoolOptions;
     protected useSessionPool: boolean;
     protected crawlingContexts = new Map<string, Context>();
-    protected isRunningPromise?: Promise<void>;
     protected autoscaledPoolOptions: AutoscaledPoolOptions;
     protected events: EventManager;
 
@@ -404,11 +404,6 @@ export class BasicCrawler<
             handleFailedRequestFunction,
             failedRequestHandler,
         } = options;
-
-        if (!requestList && !requestQueue) {
-            const msg = 'At least one of the parameters "options.requestList" and "options.requestQueue" must be provided!';
-            throw new ArgumentError(msg, this.constructor);
-        }
 
         this.requestList = requestList;
         this.requestQueue = requestQueue;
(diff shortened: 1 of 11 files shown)


packages/basic-crawler/src/internals/basic-crawler.ts
```diff
@@ -527,27 +521,27 @@ export class BasicCrawler<
...
-        if (this.isRunningPromise) return this.isRunningPromise;
-
+    async run(): Promise<FinalStatistics> {
         await this._init();
-        this.isRunningPromise = this.autoscaledPool!.run();
         await this.stats.startCapturing();
 
         try {
-            await this.isRunningPromise;
+            await this.autoscaledPool!.run();
```
mnmkng, line 529: Aren't we gonna miss the promise? I remember it sometimes being useful. But maybe it was an antipattern.

packages/basic-crawler/src/internals/basic-crawler.ts
```diff
@@ -661,7 +655,7 @@ export class BasicCrawler<
      * then retries them in a case of an error, etc.
      */
     protected async _runTaskFunction() {
-        const source = this.requestQueue || this.requestList!;
+        const source = this.requestQueue || this.requestList || await this.getRequestQueue();
```
mnmkng, line 658: Is this to enable crawlers automatically using the default queue?

B4nan: Yop, we can't do this in the constructor as it needs to be awaited. Next time `_runTaskFunction` will be invoked it will end up with `this.requestQueue` being set.

packages/core/src/enqueue_links/enqueue_links.ts
```diff
@@ -212,8 +212,10 @@ export async function enqueueLinks(options: EnqueueLinksOptions): Promise<BatchA
         urlPatternObjects.push(...constructRegExpObjectsFromRegExps(regexps));
     }
 
+    options.strategy ??= urlPatternObjects.length > 0 ? undefined : EnqueueStrategy.SameSubdomain;
```
vladfrangu, line 215: Maybe wrap in an if check instead? 
```suggestion
    if (urlPatternObjects.length > 0) {
    	options.strategy ??= EnqueueStrategy.SameSubdomain;
	}
```

B4nan: you can improve on this as part of #64, i've done this only so the example e2e test works correctly

(btw my code does the opposite than your suggestion, we don't want the default to be set if there are any url patterns)

PR #76, opened 2022-04-11
fix: enqueueLinks with redirect urls

### LINKED ISSUE #64: Fix `enqueueLinks` with explicit pseudourls and redirecting urls

vladfrangu:
Closes #64


packages/basic-crawler/src/internals/basic-crawler.ts
```diff
@@ -906,12 +909,8 @@ export class BasicCrawler<
 }
 
 /** @internal */
-export async function basicCrawlerEnqueueLinks(options: BasicCrawlerEnqueueLinksOptions, requestQueue?: RequestQueue) {
-    return enqueueLinks({
-        // FIXME this does not make much sense, as the instance would be ignored by any crawler - the argument needs to be required
-        requestQueue: requestQueue ?? await RequestQueue.open(),
-        ...options,
-    });
+export async function basicCrawlerEnqueueLinks(options: EnqueueLinksOptions) {
```
B4nan, line 912: let's remove this and use `enqueueLinks` directly

test/e2e/test-cheerio-enqueue-links/test.mjs
```diff
@@ -0,0 +1,28 @@
+import { Actor } from 'apify';
+import { CheerioCrawler } from '@crawlee/cheerio';
+import deepEqual from 'deep-equal';
+import { expect } from '../tools.mjs';
+
+const crawler = new CheerioCrawler({
+    async requestHandler({ $, enqueueLinks, request, log }) {
+        const { url, loadedUrl } = request;
+
+		const pageTitle = $('title').first().text();
```
B4nan, line 10: mixed indent, we should add editorconfig file... :]

test/e2e/test-cheerio-enqueue-links/test.mjs
```diff
@@ -0,0 +1,28 @@
+import { Actor } from 'apify';
+import { CheerioCrawler } from '@crawlee/cheerio';
+import deepEqual from 'deep-equal';
+import { expect } from '../tools.mjs';
+
+const crawler = new CheerioCrawler({
```
B4nan, line 6: lets test the same with one of the browser crawlers too

packages/cheerio-crawler/src/internals/cheerio-crawler.ts
```diff
@@ -1062,23 +1065,39 @@ export class CheerioCrawler<JSONData = unknown> extends BasicCrawler<
...
+    if (options?.baseUrl) {
+        // User provided base url takes priority
+        baseUrl = options.baseUrl;
+    } else if (options?.strategy === EnqueueStrategy.All) {
+        // We can assume users want to go off the domain in this case
+        baseUrl = finalRequestUrlOrigin ?? originalRequestUrlOrigin;
+    } else if (originalRequestUrlOrigin === finalRequestUrlOrigin) {
+        // If the domains are the same, we can use the original domain
+        baseUrl = originalRequestUrlOrigin;
+    }
```
B4nan, line 1090: the code looks almost identical to the one in browser crawler, can we refactor this out? i guess this version would work for the browser crawler too, right?

packages/cheerio-crawler/src/internals/cheerio-crawler.ts
```diff
@@ -1062,23 +1065,39 @@ export class CheerioCrawler<JSONData = unknown> extends BasicCrawler<
...
+    // else {
+    // TODO: Do we want to use finalRequestUrlOrigin if domains match or leave it as undefined?
```
B4nan, line 1092: not sure about this, can you give me an example on some edge cases where this matters?

B4nan (changes requested):

packages/cheerio-crawler/src/internals/cheerio-crawler.ts
vladfrangu: The only time I could think this would be an edge case is `domain.com/redirect` -> `subdomain.domain.com` but I'm not too sure... That would not have a baseUrl set so no links would be enqueued

test/e2e/test-cheerio-enqueue-links/test.mjs
B4nan: yeah maybe do it separately, we can merge that right ahead

packages/basic-crawler/src/internals/basic-crawler.ts
vladfrangu: For basic crawlers only? I agree, but for the rest I think they should stay separate (+ makes testing easier)

packages/cheerio-crawler/src/internals/cheerio-crawler.ts
B4nan: ok, will leave it up to you, i guess we dont need to handle all edge cases right now, especially when we are not sure they exist. we will catch issues like this when we start testing on platform and in real projects.

packages/basic-crawler/src/internals/basic-crawler.ts
B4nan: i would inline the usage of `basicCrawlerEnqueueLinks`, rest can stay the same (although i would like to not export those directly, but we can deal with that later, there are more candidates that could live in some `internals.js` barrel that would be listed in package exports, not won't be documented)

packages/cheerio-crawler/src/internals/cheerio-crawler.ts
vladfrangu: Should we leave the TODO for now so we don't forget or clean it up and handle it when issues actually arise?

B4nan: i would either keep it uncommented or remove it fully (together with the unused dependency, unless its used elsewhere too) 

maybe better to have the baseUrl inferred all the time, i am sure we will catch it early if it would be problematic, so i'd be in favour of uncommenting

packages/core/src/enqueue_links/enqueue_links.ts
```diff
@@ -262,3 +262,45 @@ export async function enqueueLinks(options: EnqueueLinksOptions): Promise<BatchA
 
     return requestQueue.addRequests(requests);
 }
+
+export function resolveBaseUrl({
```
B4nan, line 266: let's mark this as internal, otherwise ready to merge

szmarczak (approved):

B4nan (approved):

PR #77, opened 2022-04-13
feat(AutoscalePool): add support for max tasks per minute

### LINKED ISSUE #66: Autoscaling should enable max requests per minute setting
**Describe the feature**
Ability to set up `maxRequestsPerMinute` or similar option in the `autoscaledPoolOptions`.

**Motivation**
Concurrency cannot really be used to control how many requests will hit the target website because it depends on if the request takes 0.1 sec or 10 sec. This is a nice feature to handle blocking or not overloading target website.

**Constraints**
How it will work with other limits, I guess it will scale down either on concurrency or RPM, just depending on what comes first. But not sure how we would include it somehow to the metrics, etc.

cc @petrpatek if you have more ideas

vladfrangu:
Closes #66

### DIFF
--- a/packages/core/src/autoscaling/autoscaled_pool.ts
+++ b/packages/core/src/autoscaling/autoscaled_pool.ts
@@ -203,8 +203,8 @@ export class AutoscaledPool {
     private maybeRunInterval!: BetterIntervalID;
     private queryingIsTaskReady!: boolean;
     private queryingIsFinished!: boolean;
-    private _currentTasksDoneThisMinute = 0;
-    private minuteCounterResetInterval?: BetterIntervalID;
+    private tasksDonePerSecondInterval?: BetterIntervalID;
+    private _tasksPerMinute: number[] = Array.from({ length: 60 }, () => 0);
 
     constructor(
         options: AutoscaledPoolOptions,
@@ -276,7 +276,7 @@ export class AutoscaledPool {
         this.reject = null;
         this._autoscale = this._autoscale.bind(this);
         this._maybeRunTask = this._maybeRunTask.bind(this);
-        this._resetMinuteCounter = this._resetMinuteCounter.bind(this);
+        this._incrementTasksDonePerSecond = this._incrementTasksDonePerSecond.bind(this);
 
         // Create instances with correct options.
         const ssoCopy = { ...systemStatusOptions };
@@ -370,7 +370,7 @@ export class AutoscaledPool {
 
         if (this.maxTasksPerMinute !== Infinity) {
             // Start the interval that resets the counter of tasks per minute.
-            this.minuteCounterResetInterval = betterSetInterval(this._resetMinuteCounter, 60_000);
+            this.tasksDonePerSecondInterval = betterSetInterval(this._incrementTasksDonePerSecond, 1000);
         }
 
         try {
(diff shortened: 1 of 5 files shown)


B4nan:
cc @mnmkng @pocesar

B4nan (approved):
lgtm

vladfrangu:
From the original issue it seems to be a "no matter the concurrency, stop if it reaches this many tasks in that minute" (at least that's how I understood it and implemented it), but it's possible I could be wrong

mnmkng:
Yeah, the original issue is fine and you're correct. It's just the counting of the minute that's incorrect IMO. You reset an interval once a minute and then add all the requests inside this interval. But it should not have an interval, it needs to track the time of tasks and then look back 60 seconds and count how many tasks.

An example, let's say the limit is 50.

We start the interval, nothing happens for 55 seconds, then the user starts 50 tasks. 5 secs later the limit kicks in, resets the counter and the user can immediately start 50 tasks again, running 100 tasks in 6 seconds.

The correct way to approach it is to:

1. when a user is starting a task, note the time T
2. find all the tasks which were started between T and T-60
3. count them, if the count is over limit, prevent starting task

Does that make sense?

vladfrangu:
@mnmkng I think I moved it to a rolling interval, can you take a peek? 🙏

mnmkng (approved):

packages/core/src/autoscaling/autoscaled_pool.ts
```diff
@@ -677,9 +677,7 @@ export class AutoscaledPool {
     protected async _incrementTasksDonePerSecond(intervalCallback: () => void) {
         this._tasksPerMinute.unshift(0);
 
-        if (this._tasksPerMinute.length > 60) {
-            this._tasksPerMinute.pop();
-        }
+        this._tasksPerMinute.pop();
```
mnmkng, line 680: Haha, I had a comment half written that the if was useless, but in the end I was like "it doesn't hurt anything" 😄

packages/core/src/autoscaling/autoscaled_pool.ts
```diff
@@ -640,6 +670,19 @@ export class AutoscaledPool {
 
         betterClearInterval(this.autoscaleInterval);
         betterClearInterval(this.maybeRunInterval);
+        if (this.tasksDonePerSecondInterval) betterClearInterval(this.tasksDonePerSecondInterval);
         if (this.snapshotter) await this.snapshotter.stop();
     }
+
+    protected async _incrementTasksDonePerSecond(intervalCallback: () => void) {
```
pocesar, line 677: seems there's no need for this to be `async`, even the `intervalCallback` is synchronous 🙂

pocesar:
I'd just point out that intervals are "best effort" and can be stuck if the event loop is busy, and can happen quite frequently on lower memory settings on the platform (extreme example: 0.03125 vCPU and 128MB, but seen happening on 1024MB as well), as they will be fighting: io, synchronous code and scheduled work, only one will win (hint, it's not the timers 😄)

I'm still sticking to my guns and trying to incentivize the rework on SDK 3 to have a better approach instead of porting the current way to the next major version

metalwarrior665 (commented):
Thanks :)

mnmkng:
Yeah, as Paulo says, doing this without intervals would be better, but this is good enough. We'll improve it if needed. It doesn't matter that much if the timer gets stuck for a few ms and lets one or two more tasks run.

pocesar:
@mnmkng you couldn't be more wrong, sorry. even locally I can make the timers and console.log calls hang for 1-2 minutes with simple parsing of a big sitemap

mnmkng:
@pocesar when you parse a sitemap for 2 mins, it doesn't matter too much if the timers get stuck, because everything gets stuck. As I said before, you're right. But in this case the real world impact of the stuck timers will be negligible.

B4nan:
We agreed with @vladfrangu we will merge this as is, and we will be happy to revisit if @pocesar (or anybody else) can provide some way to break it. We want to have some initial beta published on NPM within next few weeks, so I hope it will be easier to reason about things like this after we do some real world testing.

PR #82, opened 2022-04-21
feat: ProxyConfiguration.newUrlFunction made async

### LINKED ISSUE #80: Awaiting for proxyConfiguration.newProxyInfo
**Describe the feature**
Proposal to add await to newProxyInfo in CheerioCrawler and BrowserCrawler, here:
https://github.com/apify/apify-js/blob/d44f3b506822467fd0ec518a1436513207cd56c8/src/crawlers/cheerio_crawler.js#L475
https://github.com/apify/apify-js/blob/d44f3b506822467fd0ec518a1436513207cd56c8/src/crawlers/browser_crawler.js#L508

**Motivation**
That will allow to use http calls in newUrlFunction.

E.g. we use external proxy manager server which distributes proxies to Actors, but that currently only works by bulk preloading of proxies from that server by Actor and making newUrlFunction get proxies from that preloaded pool. 

Would be great to just dynamically get proxy on each call of the newUrlFunction, which is currently impossible as there's no await in those two places listed above.

**Constraints**
Addition of the await in both of the places presumably wouldn't break any of the existing code.

barjin:
Closes #80

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 - some `AutoscaledPool` options are no longer available:
   - `cpuSnapshotIntervalSecs` and `memorySnapshotIntervalSecs` has been replaced with top level `systemInfoIntervalMillis` configuration
   - `maxUsedCpuRatio` has been moved to the top level configuration
+  - `ProxyConfiguration.newUrlFunction` can be async. `.newUrl()` and `.newProxyInfo()` now return promises.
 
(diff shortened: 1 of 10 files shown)


B4nan (approved):
looking good, please add it to the changelog, as this is a breaking change, both `newProxyInfo` and `newUrl` are public API users might be using directly

we should also add some tests that use async proxy provider

PR #83, opened 2022-04-21
chore: convert e2e tests to using the crawlers directly

### LINKED ISSUE #81: Rewrite E2E tests to use the crawlers directly instead of apify actors
Currently the E2E tests are just using the actors from https://github.com/apify/apify-ts/tree/master/packages/actor-scraper. This package will probably live in a different monorepo (together with actor sdk, apify client, apify cli, ...), and for the purpose of this monorepo, we should rather test the crawler classes directly. 

An example migration can be seen here: https://github.com/apify/apify-ts/blob/master/test/e2e/test-che-default/test.mjs

Please adjust the rest of the tests this way, so they define the crawler directly instead of using apify scrapers. Note that some of the tests might be failing due to redesigned things on apify.com. If it's something easy to solve, please fix it, if not, add a TODO to the test - once ready, we will migrate the tests (as well as the getting started guide in docs) to use the [demo store](https://demo-webstore.apify.org/) instead of apify.com, so we don't have to deal with this again and again. (right now the demo store is too simple, has no categories, too few products, variants, etc.)

Maybe we could also implement a way to skip such tests, could be a simple `process.exit()` with some "skip test exit code" wrapped in a helper function.

vladfrangu:
Closes #81


AndreyBykov:
I am probably a bit out of context here, but I just realized that we're currently changing e2e to basically normal actors, while originally these were testing the crawlers (public actor crawlers, i.e. cheerio, puppeteer, and web scraper), including their input. This applies to my PR too btw. Now we basically test the SDK here, and that's ok, but just double-checking whether it's something that we do intentionally as at some point we would need to upgrade the scrapers (public actors) too @B4nan

As for our discussion on Slack regarding SSL tests @vladfrangu - I tried locally and I think they work as expected, one test ignores SSL errors, and the other one doesn't. Need to adjust the numbers in assertions and it should be good.

There are web-scraper tests too, but given we are converting these to e2e SDK test - I guess we could just remove web-scraper ones for now?

B4nan:
> Now we basically test the SDK here

Yes, e2e tests of the SDK, on the platfom, via apify-client rather than storage-local. 

This repository will not contain the apify package, neither apify storage local or the actor scraper repository - those will go to another monorepo. We could have such tests for the public scrapers there, but not here, here we care about the SDK/crawlee code only.

AndreyBykov:
In this case, I guess we could remove all `web-scraper` tests straight away (not to be distracted by them here), as it's also using puppeteer under the hood.

B4nan:
Yeah exactly, we already discussed that with @vladfrangu, I believe that's why the PR is marked as draft, other than that it should be ready.

AndreyBykov:
Got it, cool! One more thing - Since In my draft I will basically have it now with a full actor directory (with Docker file) - wanted to see how we're going to proceed? Vlad could probably merge it and I will convert it as I did with the first test + double-check the assertions and stuff, or should just merge the PRs? Just checking what and when should I do with my part :)

B4nan:
Yop, we will first merge this one, and you will convert them all to the new shape.

packages/playwright-crawler/src/internals/playwright-crawler.ts
```diff
@@ -250,4 +253,18 @@ export class PlaywrightCrawler extends BrowserCrawler<{ browserPlugins: [Playwri
 
         return gotoExtended(crawlingContext.page, crawlingContext.request, gotoOptions);
     }
+
+    protected override async _applyCookies({ session, request, page }: PlaywrightCrawlContext, preHooksCookies: string, postHooksCookies: string) {
+        const sessionCookie = session?.getPuppeteerCookies(request.url) ?? [];
```
B4nan, line 258: maybe we should rename the `getPuppeteerCookies` method? also looking at the code, we depend on puppeteer types in core because of it, we will need to fix that too

(let's not deal with that in this PR, just a thought)

packages/cheerio-crawler/src/internals/cheerio-crawler.ts
```diff
@@ -794,16 +793,19 @@ export class CheerioCrawler<JSONData = unknown> extends BasicCrawler<
 
         const gotOptions = {} as OptionsInit;
 
-        if (this.useSessionPool) {
-            this._applySessionCookie(crawlingContext, gotOptions);
-        }
-
         const { request, session } = crawlingContext;
-        const cookieSnapshot = request.headers?.Cookie ?? request.headers?.cookie;
+        const preNavigationHooksCookies = request.headers?.Cookie ?? request.headers?.cookie ?? '';
```
B4nan, line 797: it feels like we are introducing quite a lot of code duplication, could we try to move at least parts of this to basic crawler?

packages/puppeteer-crawler/src/internals/puppeteer-crawler.ts
```diff
@@ -4,13 +4,15 @@ import {
...
     BrowserHook,
+    PuppeteerCookie,
 } from '@crawlee/browser';
 import {
     Dictionary,
 } from '@crawlee/utils';
 import { BrowserPoolOptions, PuppeteerPlugin } from '@crawlee/browser-pool';
 import ow from 'ow';
 import { HTTPResponse, LaunchOptions, Page } from 'puppeteer';
+import { Cookie } from 'tough-cookie';
```
B4nan, line 15: undeclared dependency

packages/cheerio-crawler/src/internals/cheerio-crawler.ts
vladfrangu: We kind of can't... each extended crawler overrides how the hooks are called / where they're called, so we need to have this. We could, however, make the `request.headers?.Cookie ?? request.headers?.cookie ?? ''` a basic crawler method to reduce that code logic dupe

packages/puppeteer-crawler/src/internals/puppeteer-crawler.ts
vladfrangu: It comes from core, but I can explicitly add it too if needed 👍

packages/playwright-crawler/src/internals/playwright-crawler.ts
vladfrangu: Opened an issue for this so we can track it for the future 👍

packages/puppeteer-crawler/src/internals/puppeteer-crawler.ts
B4nan: that's not how dependency resolution works, you are depending on an undefined behaviour that might differ across package managers (e.g. this would not work with yarn/berry with pnp enabled). dependencies you use need to be explicit.

packages/cheerio-crawler/src/internals/cheerio-crawler.ts
vladfrangu: Nevermind, remembered what I wanted to change and now code dupe is no more

packages/puppeteer-crawler/tsconfig.json
```diff
@@ -1,4 +1,4 @@
 {
   "extends": "../../tsconfig.json",
-  "include": ["src/**/*"]
+  "include": ["src/**/*", "puppeteer-launcher.ts"]
```
B4nan, line 3: why? it is under `src` folder as well as exported from the root index.ts, or not?

test/apify/crawlers/basic_browser_crawler.ts
```diff
@@ -1,4 +1,4 @@
-import { BrowserCrawler, PuppeteerCrawlContext, PuppeteerCrawlerOptions } from '@crawlee/puppeteer';
+import { BrowserCrawler, PuppeteerCookie, PuppeteerCrawlContext, PuppeteerCrawlerOptions } from '@crawlee/puppeteer';
```
B4nan, line 1: seems to be unused, given there are no other changes in this file

test/e2e/test-che-initial-cookies/test.mjs
```diff
@@ -1,62 +1,64 @@
...
-        {
-            name: 'store',
-            value: 'value store',
-        },
-        {
-            name: 'market_place',
-            value: 'value market place',
-        },
-    ],
+    useSessionPool: true,
```
B4nan, line 55: `useSessionPool` defaults to true, no need to specify this explicitly

test/e2e/test-che-page-info/test.mjs
```diff
@@ -1,55 +1,49 @@
...
-            await skipLinks();
 
             const uniqueIdentifier = url.split('/').slice(-2).join('/');
             const title = $('header h1').text();
             const description = $('header span.actor-description').text();
             const modifiedDate = $('ul.ActorHeader-stats time').attr('datetime');
             const runCount = $('ul.ActorHeader-stats > li:nth-of-type(3)').text().match(/[\d,]+/)[0].replace(/,/g, '');
 
-            return {
+            await pushData({
```
B4nan, line 31: interesting, there should be no `pushData` helper i'd say, only `Actor.pushData`

will remove the function myself, please use `Actor.pushData` here so things won't break because of it

(same for `openDataset`)

https://github.com/apify/apify-ts/commit/8ef77f54e0bf40a0e750fc78724e8dbc12f8bc14

test/e2e/test-che-ssl-errors-disabled/test.mjs
```diff
@@ -1,46 +1,40 @@
...
+            });
         }
 
-        async function handleDetail({ request, log, $ }) {
+        async function handleDetail() {
             const { url } = request;
             log.info(`Scraping ${url}`);
             const title = $('title').text();
-            return { url, title };
+            await pushData({ url, title });
```
B4nan, line 29: same here, use `Actor.pushData`

test/e2e/test-che-ssl-errors/test.mjs
```diff
@@ -1,45 +1,40 @@
...
+            });
         }
 
-        async function handleDetail({ request, log, $ }) {
+        async function handleDetail() {
             const { url } = request;
             log.info(`Scraping ${url}`);
             const title = $('title').text();
-            return { url, title };
+            await pushData({ url, title });
```
B4nan, line 29: dtto

B4nan (approved):
lgtm, good job! let's fix the `pushData` and `useSessionPool` things, otherwise ready to merge

packages/puppeteer-crawler/tsconfig.json
vladfrangu: I accidentally moved the launcher file in VSC by mistake and it updated imports automatically 😅 Nice catch

vladfrangu:
That should be everything :D

test/e2e/test-che-default/test.mjs
```diff
@@ -1,6 +1,8 @@
-import { Actor } from 'apify';
+import { Actor, pushData } from 'apify';
```
B4nan, line 1: does this really work? it should not, its no longer an exported helper, its a static method of `Actor` class

B4nan: seems to be fixed in all places except this one? :]

B4nan (approved):

B4nan:
perfect, lets wait for the green light :]

PR #93, opened 2022-04-29
feat: Request.skipNavigation

### LINKED ISSUE #79: Handle static requests outside of the browser's page
**Describe the feature**
While using a BrowserCrawler (Playwright or Puppeteer for ex), run the static requests outside of the browser (via a fetch or got-scraping) instead of using a page (by overriding its settings).

**Motivation**
Within a single actor I want to run static requests while still benefitting from some features like:
- the scaling (concurrency against available resources)
- the general proxy settings
- the retries settings
- the requests states (for general tracking & resurrections capabilities)

I don't like doing requests within the **requestHandleFunction**  while we already have a requestQueue that should be here to handle these (with many benefits).

**Constraints**
It needs to use all settings passed to the browser (session, proxy, cookies, etc).

**Proof of Concept**
Available here, I've made a quick patch that does what I described above. Also it doesn't handle some aspects (like sessions) for now, let's say it does the trick.

barjin:
closes #79 

Intended usage:
```javascript
const r = new Request({
    url: 'https://example.com/',
    skipNavigation: true,
});

...

const crawler = new WhateverCrawler({
    requestQueue,
    requestHandler: async ({request, page}) => {
        if(request.skipNavigation){
            console.log("this is a custom request!");
            return;
        }
        console.log("This is a regular request.");
    },
});
```
The serialization of `Request.internalVariables` in the `RequestQueue` constructor using a Proxy client object feels a bit (a lot) out of the place. Also, the new structure of the `Request` object breaks tests (probably expected).

Ideas welcome!

### DIFF
--- a/packages/basic-crawler/src/internals/basic-crawler.ts
+++ b/packages/basic-crawler/src/internals/basic-crawler.ts
@@ -725,7 +725,9 @@ export class BasicCrawler<
 
         try {
             await addTimeoutToPromise(
-                () => this._runRequestHandler(crawlingContext),
+                crawlingContext.request.skipNavigation
+                    ? (() => this.requestHandler(crawlingContext) as any)
+                    : () => this._runRequestHandler(crawlingContext),
                 this.requestHandlerTimeoutMillis,
                 `handleRequestFunction timed out after ${this.requestHandlerTimeoutMillis / 1000} seconds (${request.id}).`,
             );
--- a/packages/basic-crawler/src/internals/basic-crawler.ts
+++ b/packages/basic-crawler/src/internals/basic-crawler.ts
@@ -725,9 +725,7 @@ export class BasicCrawler<
 
         try {
             await addTimeoutToPromise(
-                crawlingContext.request.skipNavigation
-                    ? (() => this.requestHandler(crawlingContext) as any)
-                    : () => this._runRequestHandler(crawlingContext),
+                () => this._runRequestHandler(crawlingContext),
                 this.requestHandlerTimeoutMillis,
                 `handleRequestFunction timed out after ${this.requestHandlerTimeoutMillis / 1000} seconds (${request.id}).`,
             );
--- a/packages/core/src/request.ts
+++ b/packages/core/src/request.ts
@@ -95,7 +95,7 @@ export class Request {
     headers?: Record<string, string>;
 
     /** Custom user data assigned to the request. */
-    userData: Record<string, unknown> & { '__crawlee'?: Record<string, unknown> };
+    userData: Record<string, any>;
 
     /**
      * ISO datetime string that indicates the time when the request has been processed.
@@ -164,30 +164,28 @@ export class Request {
         this.userData = { ...userData };
         this.handledAt = handledAt as unknown instanceof Date ? (handledAt as Date).toISOString() : handledAt!;
 
-        Object.defineProperty(this.userData, '__crawlee', {
-            // eslint-disable-next-line no-underscore-dangle
-            value: this.userData.__crawlee,
-            enumerable: false,
-            writable: true,
-        });
-
-        Object.defineProperty(this.userData, 'toJSON', {
-            value: () => {
-                return {
+        Object.defineProperties(this.userData, {
+            __crawlee: {
+                // eslint-disable-next-line no-underscore-dangle
+                value: this.userData.__crawlee,
+                enumerable: false,
+                writable: true,
+            },
+            toJSON: {
+                value: () => ({
                     ...this.userData,
-                    toJSON: undefined as any,
                     // eslint-disable-next-line no-underscore-dangle
                     __crawlee: this.userData?.__crawlee,
-                };
+                }),
+                enumerable: false,
             },
-            enumerable: false,
         });
 
         if (skipNavigation != null) this.skipNavigation = skipNavigation;
     }
 
     /** Tells the crawler processing this request to skip the navigation and process the request directly. */
-    get skipNavigation() : boolean {
+    get skipNavigation(): boolean {
         // eslint-disable-next-line no-underscore-dangle
         return !!this.userData?.__crawlee?.skipNavigation ?? false;
     }
@@ -316,7 +314,7 @@ export interface RequestOptions {
      * Custom user data assigned to the request. Use this to save any request related data to the
      * request's scope, keeping them accessible on retries, failures etc.
      */
-    userData?: Record<string, unknown>;
+    userData?: Record<string, any>;
 
     /**
      * If `false` then the hash part of a URL is removed when computing the `uniqueKey` property.
--- a/packages/core/src/request.ts
+++ b/packages/core/src/request.ts
@@ -169,14 +169,13 @@ export class Request {
             value: this.userData.__crawlee,
             enumerable: false,
             writable: true,
-            configurable: true,
         });
 
         Object.defineProperty(this.userData, 'toJSON', {
             value: () => {
                 return {
                     ...this.userData,
-                    toJSON: undefined,
+                    toJSON: undefined as any,
                     // eslint-disable-next-line no-underscore-dangle
                     __crawlee: this.userData?.__crawlee,
                 };
--- a/packages/core/src/request.ts
+++ b/packages/core/src/request.ts
@@ -95,10 +95,7 @@ export class Request {
     headers?: Record<string, string>;
 
     /** Custom user data assigned to the request. */
-    userData?: Record<string, unknown>;
-
-    /** Private property for storing internal request metadata. */
-    private internalVariables: Record<string, unknown>;
+    userData: Record<string, unknown> & { '__crawlee'?: Record<string, unknown> };
 
     /**
      * ISO datetime string that indicates the time when the request has been processed.
@@ -166,20 +163,39 @@ export class Request {
         this.headers = { ...headers };
         this.userData = { ...userData };
         this.handledAt = handledAt as unknown instanceof Date ? (handledAt as Date).toISOString() : handledAt!;
-        // eslint-disable-next-line no-underscore-dangle
-        this.internalVariables = { ...(userData.__crawlee as Record<string, unknown> ?? {}) };
-        // eslint-disable-next-line no-underscore-dangle
-        delete this.userData.__crawlee;
+
+        Object.defineProperty(this.userData, '__crawlee', {
+            // eslint-disable-next-line no-underscore-dangle
+            value: this.userData.__crawlee,
+            enumerable: false,
+            writable: true,
+            configurable: true,
+        });
+
+        Object.defineProperty(this.userData, 'toJSON', {
+            value: () => {
+                return {
+                    ...this.userData,
+                    toJSON: undefined,
+                    // eslint-disable-next-line no-underscore-dangle
+                    __crawlee: this.userData?.__crawlee,
+                };
+            },
+            enumerable: false,
+        });
+
         if (skipNavigation != null) this.skipNavigation = skipNavigation;
     }
 
     /** Tells the crawler processing this request to skip the navigation and process the request directly. */
     get skipNavigation() : boolean {
-        return !!this.internalVariables.skipNavigation ?? false;
+        // eslint-disable-next-line no-underscore-dangle
+        return !!this.userData?.__crawlee?.skipNavigation ?? false;
     }
 
     set skipNavigation(value: boolean) {
-        this.internalVariables.skipNavigation = value;
+        // eslint-disable-next-line no-underscore-dangle
+        this.userData.__crawlee = { ...this.userData?.__crawlee, skipNavigation: value };
     }
 
     /**
--- a/packages/core/src/request.ts
+++ b/packages/core/src/request.ts
@@ -94,14 +94,11 @@ export class Request {
     /** Object with HTTP headers. Key is header name, value is the value. */
     headers?: Record<string, string>;
 
+    /** Private store for the custom user data assigned to the request. */
+    private _userData: Record<string, any> = {};
+
     /** Custom user data assigned to the request. */
     userData: Record<string, any> = {};
-
-    /** Private store for the custom user data assigned to the request. */
-    #userData: Record<string, any>;
-
-    /** Private store for internal `Request` variables */
-    #internalVariables: Record<string, any>;
 
     /**
      * ISO datetime string that indicates the time when the request has been processed.
@@ -167,36 +164,50 @@ export class Request {
         this.retryCount = retryCount;
         this.errorMessages = [...errorMessages];
         this.headers = { ...headers };
-        this.#userData = { ...userData };
         this.handledAt = handledAt as unknown instanceof Date ? (handledAt as Date).toISOString() : handledAt!;
 
-        // eslint-disable-next-line no-underscore-dangle
-        this.#internalVariables = userData.__crawlee ?? {};
+        Object.defineProperties(this, {
+            _userData: {
+                value: { ...userData },
+                enumerable: false,
+                writable: true,
+            },
+            userData: {
+                get: () => this._userData,
+                set: (value: Record<string, any>) => {
+                    const newValue = { ...value };
+                    Object.defineProperties(newValue, {
+                        __crawlee: {
+                            // eslint-disable-next-line no-underscore-dangle
+                            value: this._userData.__crawlee,
+                            enumerable: false,
+                        },
+                        toJSON: {
+                            // eslint-disable-next-line no-underscore-dangle
+                            value: () => ({ ...this._userData, __crawlee: this._userData.__crawlee }),
+                            enumerable: false,
+                        },
+                    });
+                    this._userData = newValue;
+                },
+                enumerable: true,
+            },
+        });
 
         if (skipNavigation != null) this.skipNavigation = skipNavigation;
-
-        Object.defineProperty(this, 'userData', {
-            get: () => ({
-                ...this.#userData,
-                toJSON: () => ({
-                    ...this.#userData,
-                    ...(Object.keys(this.#internalVariables).length ? { __crawlee: this.#internalVariables } : {}),
-                }),
-            }),
-            set: (value: Record<string, any>) => {
-                this.#userData = value;
-            },
-            enumerable: true,
-        });
     }
 
     /** Tells the crawler processing this request to skip the navigation and process the request directly. */
     get skipNavigation(): boolean {
-        return this.#internalVariables.skipNavigation ?? false;
+        // eslint-disable-next-line no-underscore-dangle
+        return this.userData.__crawlee?.skipNavigation ?? false;
     }
 
     set skipNavigation(value: boolean) {
-        this.#internalVariables.skipNavigation = value;
+        // eslint-disable-next-line no-underscore-dangle
+        if (!this.userData.__crawlee) this.userData.__crawlee = { skipNavigation: value };
+        // eslint-disable-next-line no-underscore-dangle
+        else this.userData.__crawlee.skipNavigation = value;
     }
 
     /**
--- a/packages/core/src/request.ts
+++ b/packages/core/src/request.ts
@@ -95,7 +95,13 @@ export class Request {
     headers?: Record<string, string>;
 
     /** Custom user data assigned to the request. */
-    userData: Record<string, any>;
+    userData: Record<string, any> = {};
+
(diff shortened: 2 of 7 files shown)


B4nan:
Ouch, why do you need that proxy on RQ client? I was expecting all of this to be handled directly in the `Request` class. I believe all you need is the get/set pair, and store the value directly in `userData.__crawlee`.

barjin:
> I believe all you need is the get/set pair...

That was my initial idea, however @mnmkng [seems to not](https://github.com/apify/apify-ts/issues/79#issuecomment-1110781868) want the users to see the internal data inside `Request.userData`. I am indifferent abt this, although it makes things a tad more complicated.

> Also you should not edit the `packages/storage`

Sorry abt this, understood this during the process. All the serialization is done in `RequestQueue` now.

B4nan:
Feels like you are reading too much in between the lines :] but let's wait for him to clarify. I will rather have it there instead of maintaining this proxy solution (that will btw most probably hurt performance, proxies are quite slow if they are on the hot path).

Also, we can easily define the `__crawlee` object as non enumerable, which will effectively make it hidden for the users.

packages/core/src/storages/request_queue.ts
```diff
@@ -200,10 +200,64 @@ export class RequestQueue {
...
+        const client = options.client.requestQueue(this.id, {
             clientKey: this.clientKey,
             timeoutSecs: this.timeoutSecs,
         }) as RequestQueueClient;
+
+        // dirty quick hack to serialize the Request.internalVariables into Request.userData.
+        // needs to be solved somehow else.
+        // TODO: requestQueue tests need to be fixed. they are getting confused with the internalVariables and userData shenanigans.
+        this.client = new Proxy(client, {
+            get(target, prop: keyof RequestQueueClient) {
```
mnmkng, line 212: Can't we just override the [`toJSON`](https://javascript.info/json#custom-tojson) function of `Request` to achieve this? Or add a `request.serialize()` method and call it everywhere before passing the Request to client?

If it can be done (that largely depends on how we serialize in `apify-client`), I would prefer the `toJSON` way, because it would work even when the user decides to do something with the request manually.

packages/basic-crawler/src/internals/basic-crawler.ts
```diff
@@ -725,7 +725,9 @@ export class BasicCrawler<
 
         try {
             await addTimeoutToPromise(
-                () => this._runRequestHandler(crawlingContext),
+                crawlingContext.request.skipNavigation
+                    ? (() => this.requestHandler(crawlingContext) as any)
+                    : () => this._runRequestHandler(crawlingContext),
```
mnmkng, line 730: I don't think this is correct. There are many actions that are happening in the `_runRequestHandler` functions for all the crawlers except `Basic`. All those actions need to complete. The only thing we should skip is the navigation itself = the HTTP request in Cheerio and the `page.goto` call in browsers.

For `Basic`, `skipNavigation` probably does not make sense at all, because the navigation is never handled by the crawler.

mnmkng (changes requested):

packages/basic-crawler/src/internals/basic-crawler.ts
barjin: This follows the @B4nan 's [idea](https://github.com/apify/apify-ts/issues/79#issuecomment-1110237785) from the original issue (which I like as makes sense in the OOP inheritance way, even though `BasicCrawler` itself does not navigate). 
Although I do see that the `skipNavigation` naming suggests only skipping _navigation_, it seems to me that with Cheerio/BrowserCrawler, the rest of the actions are pretty much no-op without the navigation taking place. 

I'm sure you know better though, seeing a counterexample / reason why would the users want it would be cool, if possible :)

packages/core/src/storages/request_queue.ts
barjin: I like the `toJSON` solution, combines well with the non-enumerable property definition suggested above. Although as you say, it expects `apify-client` to JSONify the data in the first place. Again, you guys are the overlords, if this is OK, it's probably the cleanest solution idea so far :)

packages/basic-crawler/src/internals/basic-crawler.ts
mnmkng: I don't see a single no-op here in `BrowserCrawler` :) The user wouldn't even have a `page` to work with. With Cheerio most do seem like a no-op, that's true. But still not all of them. E.g. the `proxyInfo` should be set, so that the user can make their own navigation through the proper proxy.

https://github.com/apify/apify-ts/blob/c7f7b13a0ca3df9595b37aa5756fbd34b6de4502/packages/browser-crawler/src/internals/browser-crawler.ts#L472-L500

barjin: Well, given the original issue name ([Handle static requests outside of the browser's page](https://github.com/apify/apify-ts/issues/79)), I was under the impression that the user would not need the... page? :) 
But it's definitely your call, will update and we can discuss that over the updated version.

packages/core/src/request.ts
```diff
@@ -161,6 +163,38 @@ export class Request {
...
+                };
+            },
+            enumerable: false,
+        });
+
+        if (skipNavigation != null) this.skipNavigation = skipNavigation;
+    }
+
+    /** Tells the crawler processing this request to skip the navigation and process the request directly. */
+    get skipNavigation() : boolean {
```
B4nan, line 190:
```suggestion
    get skipNavigation(): boolean {
```

packages/storage/src/resource_clients/request_queue.ts
```diff
@@ -230,6 +230,11 @@ export class RequestQueueClient {
         const orderNo = this._calculateOrderNo(request, forefront);
         const id = uniqueKeyToRequestId(request.uniqueKey);
         if (request.id && id !== request.id) throw new Error('Request ID does not match its uniqueKey.');
+        // @ts-ignore
+        request.userData = { ...request.userData, __crawlee: request.internalVariables };
+        // @ts-ignore
+        delete request.internalVariables;
```
B4nan, line 236: ~let's use casting instead of ts-ignore~

actually, this package should not be touched at all, i guess this is just a residue from the original version, right?

packages/basic-crawler/src/internals/basic-crawler.ts
mnmkng: Yeah, the original idea was to enable making HTTP requests from browser crawlers, but that's just one use-case out of many. And instead of supporting this one case specifically, we decided to go with a generic solution that will skip navigation and leave everything else to the user. But you're right that for the HTTP request use-case the page does not make sense.

packages/core/src/request.ts
```diff
@@ -161,6 +163,38 @@ export class Request {
...
+            value: () => {
+                return {
+                    ...this.userData,
+                    toJSON: undefined as any,
+                    // eslint-disable-next-line no-underscore-dangle
+                    __crawlee: this.userData?.__crawlee,
+                };
+            },
+            enumerable: false,
+        });
```
B4nan, line 184: let's use a single `Object.defineProperties` call

packages/core/src/request.ts
```diff
@@ -94,7 +95,7 @@ export class Request {
     headers?: Record<string, string>;
 
     /** Custom user data assigned to the request. */
-    userData?: Record<string, unknown>;
+    userData: Record<string, unknown> & { '__crawlee'?: Record<string, unknown> };
```
B4nan, line 98: would be better to keep the old type and use casting where needed instead of publicly mentioning the `__crawlee` property, given we want users to not know about this implementation detail

i would be fine with `Record<string, any>` too, which won't require any casting

packages/core/src/request.ts
```diff
@@ -161,6 +163,38 @@ export class Request {
...
+        Object.defineProperty(this.userData, '__crawlee', {
+            // eslint-disable-next-line no-underscore-dangle
+            value: this.userData.__crawlee,
+            enumerable: false,
+            writable: true,
+        });
+
+        Object.defineProperty(this.userData, 'toJSON', {
+            value: () => {
+                return {
```
B4nan, line 176: no need for nesting

```suggestion
            value: () => ({
```

packages/core/src/request.ts
```diff
@@ -161,6 +163,38 @@ export class Request {
...
+                    ...this.userData,
+                    toJSON: undefined as any,
```
B4nan, line 178: why?

barjin: Brain fart from the past, implemented `toJSON` using `JSON.stringify()` (and tried to fix the recursion this way 🥴).

packages/core/src/request.ts
```diff
@@ -161,6 +163,36 @@ export class Request {
...
         this.handledAt = handledAt as unknown instanceof Date ? (handledAt as Date).toISOString() : handledAt!;
+
+        Object.defineProperties(this.userData, {
+            __crawlee: {
+                // eslint-disable-next-line no-underscore-dangle
+                value: this.userData.__crawlee,
+                enumerable: false,
+                writable: true,
+            },
+            toJSON: {
```
B4nan, line 174: actually, why do we need this? to me this feels like a no-op, you are adding `toJSON` to an object, and keeping the exact same behaviour as if it is not there. i really dont think we need more than just the `get/set` pair on `Request` object (plus the `__crawlee` property defined above)

B4nan: if the purpose is to hide things in logs, we should use `inspect.custom` instead, that way it will also have effect on `console.log` calls

barjin: @mnmkng proposed this solution above, since `apify-client` seems to use `JSON.stringify` when serializing data for the API request to the platform (afaiu). I don't think this `toJSON` is a no-op, as non-enumerable properties don't make it into the `JSON.stringify` output?

B4nan: oh right, i thought the purpose is the exact opposite

B4nan (approved):

packages/core/src/request.ts
mnmkng: > since apify-client seems to use JSON.stringify when serializing data for the API request to the platform (afaiu)

That would be nice to know for sure 😄 I don't remember exactly. That's why I said that it will most likely work. Have you tested it?

packages/core/src/request.ts
```diff
@@ -161,6 +163,36 @@ export class Request {
...
         this.userData = { ...userData };
         this.handledAt = handledAt as unknown instanceof Date ? (handledAt as Date).toISOString() : handledAt!;
+
+        Object.defineProperties(this.userData, {
+            __crawlee: {
+                // eslint-disable-next-line no-underscore-dangle
+                value: this.userData.__crawlee,
+                enumerable: false,
+                writable: true,
+            },
```
mnmkng, line 173: I think I imagined that the `__crawlee` property would not be there at all until we serialize the request with toJSON, but maybe this is useful for something? @B4nan

mnmkng: Oh, I see you discussed it in the other thread. Well, as I said, I thought that we would only add the property when serializing with toJSON, but I guess I don't mind that it's there when it's not enumerable.

packages/core/src/request.ts
barjin: Not sure abt the package structure (and local vs platform behaviour), but when running the debugger locally, I got [here](https://github.com/apify/apify-storage-local-js/blob/6083615fa92987f880e19755fb7d9a3818355e4d/src/resource_clients/request_queue.ts#L234), which works as expected :)

mnmkng: Yeah, that's local emulation of client. We need to check the real client as well.

B4nan:
let's add some tests, ideally an e2e test so we can see how it actually works :]

packages/core/src/request.ts
B4nan: We could move this to the setter, so we have the value only when needed.

packages/core/src/request.ts
barjin: In `apify-client-js`, the request is passed in [addRequest](https://github.com/apify/apify-client-js/blob/9ff8029f02640ed61f6a93d7615c2578f8775e50/src/resource_clients/request_queue.ts#L97) -> [httpClient.call](https://github.com/apify/apify-client-js/blob/9ff8029f02640ed61f6a93d7615c2578f8775e50/src/http_client.ts#L124) -> [httpClient._createRequestHandler](https://github.com/apify/apify-client-js/blob/9ff8029f02640ed61f6a93d7615c2578f8775e50/src/http_client.ts#L138) to `axios.request()` (in the `data` field), which seems to respect the `toJSON` overrides. This can be further ensured by serializing the data before passing them to `axios`, but might not be necessary.

packages/core/src/request.ts
mnmkng: Thinking about it now, I'm not sure if setting `toJSON` on the `userData` object is correct. What if the user does the following. I'm pretty sure there will be code like that in our actors.

```js
request.userData = { foo: "bar" }
```

I think we need to add the toJSON function directly to the Request class. @barjin

barjin: This is definitely true! The proposed solution however breaks [this line](https://github.com/apify/apify-storage-local-js/blob/6083615fa92987f880e19755fb7d9a3818355e4d/src/resource_clients/request_queue.ts#L234) in `apify-storage-local`, as the Request instance is being spread - the `toJSON` property gets spread as well and overshadows the new `id` field. 

Setter/getter pair for `userData` would not work either, as that would require custom `Request.toJSON()` for the set/get value to appear in the serialized data as well. 

Perhaps we should extend the Request interface and use the Request class more rigidly throughout the packages (no more spreads etc.)?

B4nan: > Setter/getter pair for userData would not work either, as that would require custom Request.toJSON() for the set/get value to appear in the serialized data as well.

Just mark it explicitly as `enumerable: true`, it will get serialized correctly.

https://www.typescriptlang.org/play?#code/MYewdgzgLgBAJgQygmBeGBvAvjBFdgCeAsAFCiSwhqY54EmlkDyARgFYCmwUAdHJwBmASzCcACgCcQAB06SohABQgANDADkgkCA3qMZGEZgBzTlABcMJQEo0APnhIEvAG4IANqsPGI5q0qudqiOiMhunjSu3qTGMJxgAK4AtvIIrB6cVlCSiZwxWDZkZBQQIJm8HiAmKuoAUgDKzAByvNCSoibCgsogNkVMpCC82tToAIwATADMxeTgZRVVNWowjS1tOZ3dvf1AA

packages/core/src/request.ts
```diff
@@ -95,7 +95,13 @@ export class Request {
     headers?: Record<string, string>;
 
     /** Custom user data assigned to the request. */
-    userData: Record<string, any>;
+    userData: Record<string, any> = {};
+
+    /** Private store for the custom user data assigned to the request. */
+    #userData: Record<string, any>;
```
B4nan, line 101: i would rather not use native private properties, they are very slow in node < 18 (and working with request class is indeed a hot path for us), as well as have some pitfalls (e.g. you can't create instance of such class via `Object.create()`)

also, what is the monivation for those extra properties? i would rather keep the previous solution if this is just about "hiding the __crawlee key", as that can be easily done by making it non enumerable, and we will have much simpler and more performant solution

packages/core/src/request.ts
```diff
@@ -161,38 +167,36 @@ export class Request {
...
-                    ...this.userData,
-                    // eslint-disable-next-line no-underscore-dangle
-                    __crawlee: this.userData?.__crawlee,
+        // eslint-disable-next-line no-underscore-dangle
+        this.#internalVariables = userData.__crawlee ?? {};
+
+        if (skipNavigation != null) this.skipNavigation = skipNavigation;
+
+        Object.defineProperty(this, 'userData', {
+            get: () => ({
```
B4nan, line 179: this feels also weird and slow, reading of the property will be much more common than writing, we need to optimize for reading, so we can't create fresh objects every time we read.

packages/core/src/request.ts
barjin: This was a lot of trial and error (more error than trial), but non-enumerable properties seem like a solution to many of the problems I've encountered. Big thanks for the tips! 💡

test/e2e/test-request-skipNav/test.mjs
```diff
@@ -0,0 +1,49 @@
...
+await initialize(import.meta.url);
+
+let requestCounter = 0;
+let navigationCounter = 0;
+
+// Persisting internal settings of `Request`.
+await Actor.main(async () => {
+    log.setLevel(log.LEVELS.DEBUG);
+
+    const r1 = await new Request({
```
B4nan, line 14: i guess we dont need to await the constructor, right? :]

test/e2e/test-request-skipNav/test.mjs
```diff
@@ -0,0 +1,56 @@
...
+import { initialize, expect } from '../tools.mjs';
+
+await initialize(import.meta.url);
+
+let requestCounter = 0;
+let navigationCounter = 0;
+
+// Persisting internal settings of `Request`.
+await Actor.main(async () => {
+    await purgeLocalStorage();
```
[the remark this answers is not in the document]
mnmkng: Purge should be the default in v3, yeah. I would add it to `init` as well, since `main` should be just syntax sugar for `init` and `exit` so the behavior should be exactly the same

packages/core/src/request.ts
```diff
@@ -159,8 +164,50 @@ export class Request {
...
         this.handledAt = handledAt as unknown instanceof Date ? (handledAt as Date).toISOString() : handledAt!;
+
+        Object.defineProperties(this, {
+            _userData: {
+                value: { ...userData },
+                enumerable: false,
+                writable: true,
+            },
+            userData: {
+                get: () => this._userData,
```
mnmkng, line 176: I think it's a bit weird that we only define the `__crawlee` property after either the user sets `userData` or we set `skipNavigation`. Maybe I'm reading it incorrectly, but I think it works like this:

```js
const r = new Request({url: "https://example.com"});
r.userData.__crawlee // undefined
r.userData.foo = 'bar' // still undefined
r.userData = { foo: 'bar' } // now crawlee is defined
```

or

```js
const r1 = new Request({url: "https://example.com"}); // crawlee is undefined
const r2 = new Request({url: "https://example.com", skipNavigation: true }); // here it's defined
r1.skipNavigation = true; // crawlee gets defined
```

It seems kinda random to me. I was thinking that maybe we could always have at least an empty object there, but I could be missing something.

packages/core/src/request.ts
```diff
@@ -159,8 +164,50 @@ export class Request {
         this.retryCount = retryCount;
         this.errorMessages = [...errorMessages];
         this.headers = { ...headers };
-        this.userData = { ...userData };
         this.handledAt = handledAt as unknown instanceof Date ? (handledAt as Date).toISOString() : handledAt!;
+
+        Object.defineProperties(this, {
+            _userData: {
+                value: { ...userData },
```
mnmkng, line 171: If the `userData` object comes from API with `__crawlee` defined, we will add it to `this._userData`, but won't set the property non-enumerable, so if the user then retrieves it with the `userData` getter, they will see it.

packages/core/src/request.ts
B4nan: Yes, that's how it works, its not random, its defined only once we need it. Its non enumerable and should not be used by users at all, so I don't think this really matters, but sure we can have it always defined.

If you expect we might work with it directly somewhere (or even our users), it would indeed make sense to have it always defined. But for me this is an implementation detail, not a public API.

mnmkng: I had the idea of using it for other things like stats or traces, so even if we don't implement it immediately, it would make sense to have it always defined so that we can use it internally.

packages/core/src/request.ts
B4nan: Good catch. We can either redefine the property to be non enumerable, or use the setter as you suggested.

```ts
const o = { foo: 123 };
console.log(Object.keys(o));
Object.defineProperty(o, 'foo', { enumerable: false, writable: true });
console.log(Object.keys(o));
```

packages/core/src/request.ts
B4nan: Both comments handled [here](https://github.com/apify/apify-ts/commit/79f92ecf20fc8bf10c5b5fb2f3e858c8e588e376)

PR #98, opened 2022-05-04
feat: register utils into crawling context

### LINKED ISSUE #96: Add relevant utility functions to crawling context
**Describe the feature**
Attach all the utility functions we have to the `crawlingContext` of the relevant crawlers.

```js
async handlePageFunction(crawlingContext) {
  // OLD
  await Apify.utils.enqueueLinks({ requestQueue, page });
  await Apify.utils.puppeteer.blockRequests(page);
   
  // NEW because page and requestQueue can be taken from the crawlingContext itself
  await crawlingContext.utils.enqueueLinks();
  await crawlingContext.utils.blockRequests();
}
```

Or even alternatively just:

```js
await crawlingContext.enqueueLinks();
```

**Motivation**
Now with the addition of Playwright, we have `Apify.utils.puppeteer` and `Apify.utils.playwright`. That's a bit too much, complicates code sharing and requires the user to know about those `utils` namespaces in the first place. You also need to pass parameters to the functions that could easily be auto-extracted from the crawler itself.

**Constraints**
We need to look at the functions and determine if they can be auto-curried with parameters or whether some manual transformation will be needed. There's also the question of which or how many utils can the `crawlingContext` handle and if we should just plaster them in or provide a framework to attach any future function there.

B4nan:
One of the utils is tested via e2e tests, so the place where we register things is correct and works fine.

Closes #96

### DIFF
--- a/packages/puppeteer-crawler/src/internals/utils/puppeteer_utils.ts
+++ b/packages/puppeteer-crawler/src/internals/utils/puppeteer_utils.ts
@@ -26,10 +26,11 @@ import util from 'util';
 import { LruCache } from '@apify/datastructures';
 import { Page, HTTPResponse, ResponseForRequest, HTTPRequest as PuppeteerRequest } from 'puppeteer';
 import log_ from '@apify/log';
-import { KeyValueStore, Request, validators } from '@crawlee/core';
+import { KeyValueStore, Request, storage, validators } from '@crawlee/core';
 import { Dictionary } from '@crawlee/utils';
-import { enqueueLinksByClickingElements } from '../enqueue-links/click-elements';
-import { addInterceptRequestHandler, removeInterceptRequestHandler } from './puppeteer_request_interception';
+import { enqueueLinksByClickingElements, EnqueueLinksByClickingElementsOptions } from '../enqueue-links/click-elements';
+import { addInterceptRequestHandler, InterceptHandler, removeInterceptRequestHandler } from './puppeteer_request_interception';
+import { PuppeteerCrawlingContext } from '../puppeteer-crawler';
 
 const jqueryPath = require.resolve('jquery');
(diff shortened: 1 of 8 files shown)


packages/puppeteer-crawler/src/internals/utils/puppeteer_utils.ts
```diff
@@ -603,6 +604,41 @@ export async function saveSnapshot(page: Page, options: SaveSnapshotOptions = {}
...
+        page: context.page,
+        requestQueue: context.crawler.requestQueue!,
+        ...options,
+    });
+    context.blockRequests = (options?: BlockRequestsOptions) => blockRequests(context.page, options);
+    context.blockResources = (resourceTypes?: string[]) => blockResources(context.page, resourceTypes);
+    context.cacheResponses = (cache: Dictionary<Partial<ResponseForRequest>>, responseUrlRules: (string | RegExp)[]) => {
+        return cacheResponses(context.page, cache, responseUrlRules);
+    };
+    context.compileScript = (scriptString: string, ctx?: Dictionary) => compileScript(scriptString, ctx);
```
B4nan, line 635: I've put all of them here, `compileScript` seems to be the only one that does not benefit from being bound to context, but having a single utility outside of context feels weird

only exception I haven't put here is `gotoExtended`, do we want that here too? it was used directly in the crawler, not sure if it makes sense to expose it for users this way?

B4nan:
Not sure if we want to keep the `puppeteerUtils` namespace export, given all of them will be available in the context?

mnmkng:
> Not sure if we want to keep the `puppeteerUtils` namespace export, given all of them will be available in the context?

There's a lot of usage of those utils throughout actors. I would deprecate them instead to calm the emotions 😄

PR #100, opened 2022-05-04
feat: add default memory storage

### LINKED ISSUE #91: Add in memory storage implementation
**Describe the feature**

Currently we have apify-client for the platform and `@apify/storage-local` for local development (that uses sqlite under the hood). We want to have third storage implementation, that will be in memory only (+ persisting state to JSON). This storage will reside in the crawlee monorepo, under `@crawlee/memory-storage`. 

**Motivation**

We don't want to have the default storage dependent on native modules like sqlite, as it's causing us a lot of troubles because of the need for prebuilds for various platforms.

**Constraints**

The interface needs to conform to the storage interfaces defined in the core package:

https://github.com/apify/apify-ts/blob/master/packages/core/src/storages/storage.ts

vladfrangu:
Closes #91 

So far Dataset and KVStores look to be working just like before


packages/types/src/storages.ts
```diff
@@ -230,19 +237,24 @@ export interface RequestOptions {
 }
 
 export interface RequestSchema {
-    uniqueKey: string;
+    id?: string;
     url: string;
-    method?: AllowedHttpMethods;
+    uniqueKey: string;
+    method: AllowedHttpMethods;
```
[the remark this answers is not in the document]
vladfrangu: I've undone this as the validation also allows method to be optional

vladfrangu:
I don't really understand why those two tests fail now... if I only run them they pass without issues, and even the file system structure is correct, so there must be a deadlock somewhere 😓 
If you have any tips into debugging this.. 🙏

B4nan (commented):
here is first round of comments, still have lots of files to review

packages/types/package.json
```diff
@@ -0,0 +1,54 @@
+{
+  "name": "@crawlee/types",
```
B4nan, line 2: i dont like the name much, given all it contains are the storage related interfaces, maybe something like `@crawlee/abstract-storage` could be better? or we could keep it as a namespace export and export it as `storage` from inside it. still not perfect as it exports only interfaces/types, not some abstract implementation. another idea I had is to have some shared/abstract base in `@crawlee/storage` (what we have there right now will be removed, the storage-local will stay outside of the repo and will not be part of `@crawlee` namespace), this package would 1. export the interfaces, 2. provide common ground for the implementations. so the question is, **do you see some common ground we could extract to reduce duplications to such package?** we would then reuse it in `@apify/storage-local` as well as `@crawlee/memory-storage`. i had this on my mind initially, as to me it feels weird that the storage implementation as quite complex, i would expect very simple CRUD like interface where you can plug basically any kind of storage (let it be file storage, SQL or NoSQL database, even a redis could work).

i know we also export `AllowedHttpMethods` and `Dictionary`, but both are very simple type aliases, we can afford to duplicate them if we want to (and `AllowedHttpMethods` to me sounds like something that should be part of the storage interface anyway)

packages/memory-storage/src/workers/file-storage-emulator.ts
```diff
@@ -0,0 +1,9 @@
+import { handleMessage, WorkerDirectoryMap } from './worker-utils';
+
+export class FileStorageWorkerEmulator {
```
B4nan, line 3: the filename does not reflect the class name exactly

packages/memory-storage/src/resource-clients/request-queue.ts
```diff
@@ -0,0 +1,405 @@
+import type * as storage from '@crawlee/types';
+import { s } from '@sapphire/shapeshift';
+import { randomUUID } from 'node:crypto';
+import { resolve } from 'node:path';
+import { rm } from 'node:fs/promises';
+import { move } from 'fs-extra';
```
B4nan, line 6: undeclared dependency

packages/memory-storage/src/index.ts
```diff
@@ -0,0 +1,112 @@
...
+
+export class MemoryStorage implements storage.StorageClient {
+    readonly localDataDirectory: string;
+    readonly datasetsDirectory: string;
+    readonly keyValueStoresDirectory: string;
+    readonly requestQueuesDirectory: string;
+
+    readonly keyValueStoresHandled: KeyValueStoreClient[] = []; ;
+    readonly datasetClientsHandled: DatasetClient[] = [];
+    readonly requestQueuesHandled: RequestQueueClient[] = []; ;
```
B4nan, line 33:
```suggestion
    readonly keyValueStoresHandled: KeyValueStoreClient[] = [];
    readonly datasetClientsHandled: DatasetClient[] = [];
    readonly requestQueuesHandled: RequestQueueClient[] = [];
```

packages/memory-storage/src/index.ts
```diff
@@ -0,0 +1,112 @@
...
+
+    private fileStorageWorker!: Worker | FileStorageWorkerEmulator;
```
B4nan, line 35: could the `FileStorageWorkerEmulator` maybe extend `Worker` so we dont have to do this explicitly here?

B4nan (commented):
few more comments, still rather on surface

packages/memory-storage/src/index.ts
vladfrangu: I could make `FileStorageWorkerEmulator` implement the `Worker` interface, but then most of the methods will throw/no-op. I don't know if that's a good idea vs just having this

packages/memory-storage/src/index.ts
```diff
@@ -0,0 +1,112 @@
...
+                this.createWorker();
+            });
+        } else {
+            this.fileStorageWorker = new FileStorageWorkerEmulator(directoryMap);
+        }
+    }
+
+    protected sendMessageToWorker(message: WorkerReceivedMessage) {
+        this.fileStorageWorker.postMessage(message);
+    }
```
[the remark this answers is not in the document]
vladfrangu: The method shouldn't be used by outside users, and typing it as public (even if annotated with `@internal`) will make it show up in intellisense. The bracket syntax is a trick that's by default in TS (and eslint is complaining because we're using the eslint dot-notation rule not typescript-eslints rule). The only reason I made this a method is to strictly type the message property. We *could* also make the worker be a singleton instead of created per storage instance, which would solve this. What do you think is the better option?

packages/types/package.json
vladfrangu: > do you see some common ground we could extract to reduce duplications to such package?

*If* we want to do that, we could restructure the entire `@crawlee/storage` module to act as the client interface and have options for what storage implementation it uses, which would have to extend an abstract storage class that just does CRUD operations (so then `@crawlee/memory-storage` would simply be the CRUD for memory/file system, `@apify/storage-local` would be strictly SQLite, etc).

That does sound a lot better to me (would mean any user can then create wrappers for, say, prisma, or mongodb, cockroachdb, etc.), but I don't know if that's what you're thinking about.

packages/memory-storage/src/index.ts
vladfrangu: Well technically we only really need a `postMessage` method, so we could just have an interface for that. I'll wait for your answer on this review comment, and will proceed after. 
Google Chrome Dev - 2022-05-17 at 17 56 54@2x

test/e2e/tools.mjs
```diff
@@ -22,7 +21,7 @@ export const colors = {
  * @param {string | URL} url
  */
 export function getStorage(url) {
-    return join(dirname(fileURLToPath(url)), './apify_storage');
+    return join(dirname(fileURLToPath(url)), './memory-storage');
```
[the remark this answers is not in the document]
vladfrangu: We could decide on a common name for the storage directory, and set that as the env variables (and then the storages we use in the e2e tests will go there accordingly). `2e2-test-storage`?

packages/memory-storage/src/consts.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * Length of id property of a Request instance in characters.
+ */
+export const REQUEST_ID_LENGTH = 15;
```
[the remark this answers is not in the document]
vladfrangu: Well these aren't exported at all publicly, and are only imported internally, so I don't think annotations matter much... I can add them if you'd like

packages/types/package.json
B4nan: Yes you got it right. The question is what to do with the apify client, ideally it would be also just another "adapter" to this `@crawlee/storage` package, but I guess we don't need that initally and we can just keep using the same interface. I would still keep the interfaces we have now (so defined on type level), so it works with the current apify client, and we could migrate later if we want to (or even have "apify platform storage" for crawlee and keep apify client as separate package).

B4nan:
the e2e tests are broken with this PR

https://github.com/apify/apify-ts/runs/6528531406?check_suite_focus=true

vladfrangu:
I'll need help from @AndreyBykov to figure this out... my gut says that it tries to read the JSON files before the worker actually gets to write them to disk (which would cause issues)...Adding a 10ms delay after the actor is ran for LOCAL storage implementation does seem to solve this but it feels hacky haha

B4nan:
Is it passing locally for you? How fast is it? The CI seems to get stuck or is terribly slow, I cancelled it after 75 minutes (master finishes in 30-40 min)

https://github.com/apify/apify-ts/runs/6535951717?check_suite_focus=true

Tried to rerun it and it now even failed.

https://github.com/apify/apify-ts/runs/6536335329?check_suite_focus=true

We should also stress test the storage in one e2e test with high concurrency (e.g. 100). I would maybe use the newly added `Request.skipNavigation` as we only want to test the storage, now need to do any actual requests in such test.

vladfrangu:
Found the issue... The last e2e test (request-skip-navigation) doesn't exit at the end of the test, which hangs the process running forever. I've pushed the fix 😅

vladfrangu:
https://github.com/apify/apify-ts/runs/6537945836?check_suite_focus=true 🎉

PR #103, opened 2022-05-09
fix: update purgeLocalStorage function to mimic 'purge' param of 'apify run'

### LINKED ISSUE #102: { purge: true } option of Actor.main deletes all local storages
**Describe the bug**
`{ purge: true }` option should mimic `-p` param of `apify run`, i.e. it should remove default storages besides `INPUT.json` file in default key-value store. Instead - it removes all storage (default and non-default) and all files.

**To Reproduce**
Run any actor with `{ purge: true }` param and some `INPUT.json` file or any non-default storage folder on disk.

**Expected behavior**
Only default storages should be cleared. `INPUT.json` in default key-value store should stay in place.

**System information:**
 - OS: MacOS 12.3.1
 - Node.js version: 16.14.0
 - Apify SDK version: 3.0.0-alpha.22

AndreyBykov:
This PR fixes the usage of  `{ purge: true }` param of `Actor.main()`. Currently using `{ purge: true }` will lead to the removal of everything inside of the `apify_storage` folder, while it should only remove default storage folders contents (besides `INPUT.json` file), i.e. mimic `apify run -p`.

Closes #102.

Ready for review, but do not merge it yet, also need to update the tests.

### DIFF
--- a/packages/utils/src/internals/general.ts
+++ b/packages/utils/src/internals/general.ts
@@ -1,4 +1,4 @@
-import { ENV_VARS } from '@apify/consts';
+import { ENV_VARS, LOCAL_ENV_VARS, LOCAL_STORAGE_SUBDIRS, KEY_VALUE_STORE_KEYS } from '@apify/consts';
 import contentTypeParser from 'content-type';
 import mime from 'mime-types';
 import fs from 'node:fs/promises';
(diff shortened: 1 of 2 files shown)


B4nan (approved):
maybe we should move this elsewhere (actor sdk or storage local)

or maybe not - will the memory storage use the same directories @vladfrangu?

> Ready for review, but do not merge it yet, also need to update the tests.

~How come you need to update existing tests? They somehow depend on this? Sounds weird/wrong.~

edit: oh you mean the failing tests, not e2e tests, got it

vladfrangu:
> or maybe not - will the memory storage use the same directories @vladfrangu?

Uh...memory storage does not use anything fs related. I thought that was the point of it 😅

B4nan:
During runtime yes, but once the run finishes, we need them to serialize so we can observe the state.

PR #104, opened 2022-05-10
test: move crawlee tests to `test/core` and actor sdk tests to `test/apify`

B4nan:


B4nan:
FYI @AndreyBykov @vladfrangu 

(not perfect but we need to start preparing for the split - the apify package and others will live in a different monorepo)

PR #106, opened 2022-05-15
chore: backport apify/apify-js and apify/browser-pool commits

AndreyBykov:
backport apify/apify-js and apify/browser-pool commits


packages/browser-pool/src/abstract-classes/browser-controller.ts
```diff
@@ -244,7 +244,6 @@ export abstract class BrowserController<
      * @private
      */
     // @ts-expect-error Give runtime error as well as compile time
-    // eslint-disable-next-line space-before-function-paren, @typescript-eslint/no-unused-vars
     protected abstract async _newPage(pageOptions?: NewPageOptions): Promise<NewPageResult> {
```
vladfrangu, line 247: Not an issue in this PR, but considering its all in typescript now, we can probably remove the default implementation for abstract methods in the future (to also cleanup these comments)

packages/browser-pool/src/puppeteer/puppeteer-plugin.ts
```diff
@@ -163,4 +170,11 @@ export class PuppeteerPlugin extends BrowserPlugin<typeof Puppeteer> {
         }
         */
     }
+
+    protected _isChromiumBasedBrowser(launchContext: LaunchContext<any, unknown, any, unknown, any>): boolean {
```
vladfrangu, line 174:
```suggestion
    protected _isChromiumBasedBrowser(launchContext: LaunchContext<typeof Puppeteer>): boolean {
```

packages/core/src/crawlers/statistics.ts
```diff
@@ -56,7 +55,7 @@ export class Statistics {
      */
     readonly requestRetryHistogram: number[] = [];
 
-    private keyValueStore: KeyValueStore | null = null;
+    private keyValueStore: KeyValueStore | undefined = undefined;
```
vladfrangu, line 58:
```suggestion
    private keyValueStore?: KeyValueStore = undefined;
```

AndreyBykov:
It's getting more complicated to backport stuff as things aren't exactly where they were before :)

Just wanted to double-check a few commits:

These are not back-ported - just wanted to make sure there's nothing to backport really:
https://github.com/apify/apify-js/commit/b2b5d0f9e01ac0d84c6afd5097f4da8b69db86a5
https://github.com/apify/apify-js/commit/3348175464e39b808a9bc405c0b4f2b215f9b307
https://github.com/apify/apify-js/commit/7c89d087c05eb78852efeb9b8d4cfd4b148af474
https://github.com/apify/apify-js/commit/bf16148bdc9445a50b7c4c3518009393aec3fc7e

For these wanted to be sure:
https://github.com/apify/apify-js/commit/077c31d6f5bfd51eebbb2271a9e9544e33fd956b - this one fixes utils.apifyClient early instantiation, I think there was something in PR that in version 3 it's not needed, is that correct?
https://github.com/apify/apify-js/commit/d2d5ac7476e4ddf5f5398ca24e8ca64d9a364b21 - for this one I've noticed that DEFAULT_USER_AGENT was removed, thus I assume `userAgent` part also should not be backported, right?
https://github.com/apify/apify-js/commit/e2d9233687343ea505ef2019f289bd3702a4cee6 - this one goes to puppeteer-plugin, right? I was confused for a while where to put it now :/

And for this one would just appreciate a double-check:
https://github.com/apify/apify-js/commit/898cc43fc66d951ed05911d266640ae8bf68c221

I obviously did not back port version bumps as now we're using version 3 everywhere and did not downgrade versions.

test/e2e/autoscaling-max-tasks-per-minute/actor/package.json
```diff
@@ -11,7 +11,7 @@
     },
 	"devDependencies": {
 		"@apify/eslint-config": "^0.3.1",
-		"eslint": "^8.13.0"
+		"eslint": "^8.14.0"
```
B4nan, line 14: we should remove eslint from e2e tests dependencies

vladfrangu (changes requested):
Small nits

packages/playwright-crawler/src/internals/utils/playwright-utils.ts
```diff
@@ -19,15 +19,98 @@
...
+ *
+ * @param page Playwright [`Page`](https://playwright.dev/docs/api/class-page) object.
+ * @param filePath File path
+ * @param {object} [options]
+ * @param {boolean} [options.surviveNavigations]
+ *   Enables the injected script to survive page navigations and reloads without need to be re-injected manually.
+ *   This does not mean, however, that internal state will be preserved. Just that it will be automatically
+ *   re-injected on each navigation before any other scripts get the chance to execute.
+ * @return {Promise<*>}
+ * @memberOf playwright
```
B4nan, line 57: please adjust the jsdoc
- remove `memberOf` tag
- move return type to the func declaration
- remove types from jsdoc that are explicit in the signature
- type the options parameter properly and move the jsdoc describing each option inside the interface 

same for `injectJQuery`

B4nan (approved):
looking good, just two nits

vladfrangu (approved):

vladfrangu:
I see turbo had a compile error, looks like core depends on browser-pool without explicitly declaring that in dependencies... That would be the only thing that needs to be fixed from what I see 👀

AndreyBykov:
Just for the record - you've seen my message above, right? I am mainly concerned about https://github.com/apify/apify-js/commit/d2d5ac7476e4ddf5f5398ca24e8ca64d9a364b21 - it's for playwright launcher, and DEFAULT_USER_AGENT is used there (due to some detection issues for playwright/chrome). However, DEFAULT_USER_AGENT, It kinda seems that in a case when fingerprinting is not used - we would face the same detection problem, and I should add the DEFAULT_USER_AGENT back for such cases, or maybe I am overthinking it...

AndreyBykov:
And actually `core` has `browser-pool` in `package.json` and it was added in https://github.com/apify/apify-ts/pull/73/

vladfrangu:
> https://github.com/apify/apify-js/commit/bf16148bdc9445a50b7c4c3518009393aec3fc7e

I'm not sure if we need to backport this really, we probably will need to once the docs site works again. 

> https://github.com/apify/apify-js/commit/077c31d6f5bfd51eebbb2271a9e9544e33fd956b

I don't think this is applicable anymore due to the way we handle storage now... Not 100% sure tho

> https://github.com/apify/apify-js/commit/d2d5ac7476e4ddf5f5398ca24e8ca64d9a364b21 & https://github.com/apify/apify-js/commit/e2d9233687343ea505ef2019f289bd3702a4cee6

I'm not sure about these.. I assume they do need to be backported since it can affect crawling detection? I'll leave these to @B4nan 

> https://github.com/apify/apify-js/commit/898cc43fc66d951ed05911d266640ae8bf68c221

Looks like you already backported it and it looks fine to me 👍

AndreyBykov:
https://github.com/apify/apify-js/commit/e2d9233687343ea505ef2019f289bd3702a4cee6 is backported too, I was just a bit hesitant that that `puppeteer plugin` is the right place to put it now.
https://github.com/apify/apify-js/commit/d2d5ac7476e4ddf5f5398ca24e8ca64d9a364b21 - agree that it looks like it should be backported. I am just a bit confused by the fact that `DEFAULT_USER_AGENT` was entirely removed (so the question is - is it just a coincidence and I should put it back).

vladfrangu:
> agree that it looks like it should be backported. I am just a bit confused by the fact that `DEFAULT_USER_AGENT` was entirely removed

I wonder if I missed that entirely when I split up the crawlers into separate packages...because I cannot find it anywhere 😓 

We probably need that back since I can guess it'll cause issues for crawlers once they update to crawlee

AndreyBykov:
It was there for puppeteer crawler only, then last month it was added to playwright, because there were some blocking issues, I guess I'll put it back

vladfrangu:
I'm actually lost, puppeteer is also missing the UA change... I'm gonna try checking the history, maybe I accidentally removed it, or maybe it was never here when we started moving to TS (but that wouldn't make sense as it's been on the js repo for 16months now...)

vladfrangu:
Found it, in https://github.com/apify/apify-ts/pull/85 the UA was moved to the base browser-launcher (which takes it from the LaunchContext)... I am not sure if we need to change anything then or default user agents in the crawlers now...

AndreyBykov:
I just checked with `useFingerprints: false`: Puppeteer+Chromium (both headless/headful) and headful Playwright+Chromium have `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) HeadlessChrome/101.0.4950.0 Safari/537.36` user-agent.

Headless Playwright+Chromium has `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) HeadlessChrome/101.0.4950.0 Safari/537.36` - which was exactly the problem with JS repo. I guess I am adding it back for playwright only.

AndreyBykov:
So - I've put back `DEFAULT_USER_AGENT` as there will be the same detection problem with `HeadlessChrome` user-agent otherwise. And again - would appreciate if you guys could have a look. Already sent it to Martin on slack, but for visibility:
1. I am using `...Chrome/101.0.4951.54...` as default, however - `...Chrome/101.0.4951.15...` is used on the platform. Any reason for using the version `.54` in JS repo, and not `.15`?
2. JS repo check is `if (!userAgent && (!launchOptions.executablePath || launchOptions.headless))`, I now use `if (this._isChromiumBasedBrowser() && headless && !fingerprint && !userAgent && !executablePath)` - any objections?
3. What's the point of both - assigning `launchOptions!.userAgent` and pushing user-agent to `launchOptions!.args`? Maybe I am missing something, but It feels like it's not needed - we could either use one or another. What do you think?
4. TypeScript does not like `userAgent` in `launchOptions`. But looking into fingerprinting (` const { launchOptions }: { launchOptions: any } = launchContext;` and `// @TODO: Fix the typings so they are easier to work with.` - it's a known thing, right)?

packages/core/package.json
```diff
@@ -54,6 +54,7 @@
         "@apify/log": "^1.2.0",
         "@apify/pseudo_url": "^1.1.0",
         "@apify/timeout": "^0.2.1",
+        "@crawlee/browser-pool": "^3.0.0",
```
B4nan, line 57: i dont have time to read through the comments right now, but i already said this is wrong, we need to remove the dependency, not declare it

i also dont fully understand what your problem with the user agents is, we should backport things from apify-js - there it works correctly and there you can see what is released to stable v2.

B4nan:
Regarding the `DEFAULT_USER_AGENT`, you should indeed keep as in the apify-js. we found out we had the `headless` UAs in some cases and the default agent was needed. TS version has fingerprints enabled by default - but we still want to have this working if users decide to disable fingerprints. 

I have no idea what you mean by the chrome version (`.15` vs `.54`) - this has nothing to do with platform, its a static UA fallback inside the SDK code, the value is set. if you get `.15`, its not coming from the default UA fallback but probably from fingerprint generater.

> Any reason for using the version .54 in JS repo, and not .15?

Uh, what you are asking here? :] 54 is newer, that's the reason, or am I missing something?

All the UA changes are tested, if you backport the tests and they are passing, you should be fine.

B4nan:
Actually I know why you get the different UA here - we have fingerprints enabled by default unlike in v2. So you should get a dynamic UA, not the static fallback - thats used only if you disable fingerprints.

AndreyBykov:
That's what I mean. We don't have a static fallback now. And if you disable fingerprints - the browser would be loaded with version .15, and in JS PR you also used 15 originally,  but Ondra proposed 54 and you accepted it. I know it's totally minor and we are spending too much time on it already - I just wanted to make it properly kinda, but 54 would work too of course.

B4nan:
> We don't have a static fallback now. 

How come, you should backport it the same way it is working in v2 - this needs to work the same, the only difference being v3 has fingerprints enabled by default. If we disable fingerprints, the default UA should be used. We can't use the one the browser would have on its own - its not about the exact chrome version - the default one contains "headless" in it.

AndreyBykov:
That's exactly why I had my question in the first place. Originally we have DEFAULT_USER_AGENT in consts, then it was moved to respective places (puppeteer-plugin, firefox did not have I originally I think), and then you remove DEFAULT_USER_AGENT completely... Let me check the abstract classes, but I am pretty sure it's not there either. I think somehow playwright did not need static fallback before.

But in any case - let me just double check and finish these last pieces, let's just chat in person about it once you're here, I don't want to bother you during your vacation by deeply going into how and when things happened :)

B4nan:
> I think somehow playwright did not need static fallback before.

It actually always needed this, but we never tested it. It was most probably the reason for blocking with PW + chrome.

>  let's just chat in person about it once you're here

AndreyBykov:
> I'd like to have things ready for monday... :]
I meant - chat about it post-factum, so that I would not bother you with this stuff over your holiday more, as it's not that important :)

B4nan:
> Originally we have DEFAULT_USER_AGENT in consts

Let's put it back there where it was (does not need to have its own file, can be somewhere else, but in the appropriate package, I guess `@crawlee/browser` as this is used only in browsers).

B4nan:
It's not that important where the default UA is defined, but the behaviour needs to stay the same - if fingerprints are disabled, it should be used. But I guess it's working correctly, seeing the green CI and [this test](https://github.com/apify/apify-ts/pull/106/files#diff-e043eb8adb298aa57fb5203159c02aa34d08b5ebe9deea13ad8a64ede96b3a28R91) that is exactly for that.

I can look into the browser pool dependency myself, we can merge without that - anything else I should look at or do you think its mergable?

AndreyBykov:
I'm blind, you're right, puppeteer also has headless user-agent. Updating.

AndreyBykov:
Removed `executablePath` from check for headless user-agent. For `puppeteer`/`useChrome: true` `executablePath` was present, so it still showed healdess user-agent. Moved it to browser-plugin, should cover both puppeteer/playwright (crawler/browser launch).

Should be ready for merging.

AndreyBykov:
Finally done, tests pass.

packages/playwright-crawler/src/internals/utils/playwright-utils.ts
```diff
@@ -19,15 +19,99 @@
  * @module playwrightUtils
  */
 
+import { readFile } from 'node:fs/promises';
 import ow from 'ow';
 import { Page, Response, Route } from 'playwright';
+import { LruCache } from '@apify/datastructures';
```
B4nan, line 25: undeclared dependency

B4nan (changes requested):

packages/playwright-crawler/src/internals/utils/playwright-utils.ts
AndreyBykov: Looking at the other packages - we probably need to go through them and update dependencies everywhere - e.g. puppeteer crawler does not have half of the packages there either, pretty sure other ones are worth checking too... But I guess not within this PR? Should I create an issue?

B4nan (approved):

Issue #59, opened 2022-03-14
JSDom crawler


B4nan:
- add new level of abstraction to share common parts with cheerio crawler, maybe `RequestCrawler` or `HttpCrawler`?

B4nan:
closing as implemented in crawlee via different PR

PR #107, opened 2022-05-16
feat: jsdom crawler

szmarczak:
As per sync - maybe let's move the common parts to basic crawler 🤔


packages/jsdom-crawler/src/internals/jsdom-crawler.ts
```diff
@@ -0,0 +1,1023 @@
...
+import contentTypeParser, { RequestLike, ResponseLike } from 'content-type';
+import { gotScraping, OptionsInit, Method, TimeoutError, Request as GotRequest } from 'got-scraping';
+import { IncomingHttpHeaders, IncomingMessage } from 'http';
+import iconv from 'iconv-lite';
+import ow from 'ow';
+import util from 'util';
+import { JSDOM, DOMWindow } from 'jsdom';
+
+/**
+ * Default mime types, which CheerioScraper supports.
```
szmarczak, line 34: So... I just copy-pasted the cheerio crawler and did some changes. I was thinking of renaming `cheerio` to `jsdom` everywhere but the share is just large. I'd appreciate any comment on how to solve this.

Those are basically the same crawlers with the difference that it's just `jsdom` instead of `cheerio`. However there's more to it:

JSDOM already performs encoding detection and decodes the buffer properly (just make sure we're passing a buffer), so we could remove the code that's responsible for this. For now we just read the stream and decode it ourselves, like in the cheerio crawler.

Also, JSDOM doesn't support streaming. Here's some CloudFlare's research on HTML parsers:
https://blog.cloudflare.com/html-parsing-1/
https://blog.cloudflare.com/html-parsing-2/

I don't think lack of streaming is of that important issues. In our case, more time is spent on downloading than parsing.
Although if we used WebAssembly then the parse overhead would be almost 0ms.

packages/jsdom-crawler/src/internals/jsdom-crawler.ts
```diff
@@ -0,0 +1,1023 @@
...
+const HTML_AND_XML_MIME_TYPES = ['text/html', 'text/xml', 'application/xhtml+xml', 'application/xml'];
+const APPLICATION_JSON_MIME_TYPE = 'application/json';
+const CHEERIO_OPTIMIZED_AUTOSCALED_POOL_OPTIONS = {
+    snapshotterOptions: {
+        eventLoopSnapshotIntervalSecs: 2,
+        maxBlockedMillis: 100,
+    },
+    systemStatusOptions: {
+        maxEventLoopOverloadedRatio: 0.7,
+    },
```
szmarczak, line 45: Do these need to be updated?

packages/jsdom-crawler/src/internals/jsdom-crawler.ts
```diff
@@ -0,0 +1,1023 @@
...
+            }
+
+            // It's not a JSON so it's probably some text. Get the first 100 chars of it.
+            throw new Error(`${statusCode} - Internal Server Error: ${body.substr(0, 100)}`);
+        } else if (HTML_AND_XML_MIME_TYPES.includes(type)) {
+            const body = await readStreamToString(response, encoding);
+
+            const jsdom = new JSDOM(body, {
+                url: responseStream.url,
+                contentType: type,
```
szmarczak, line 787: Note to myself: pass a cookie jar here

packages/jsdom-crawler/src/internals/jsdom-crawler.ts
```diff
@@ -0,0 +1,1023 @@
...
+            headers: { ...request.headers, ...gotOptions?.headers },
+            https: {
+                ...gotOptions?.https,
+                rejectUnauthorized: !this.ignoreSslErrors,
+            },
+            isStream: true,
+        };
+
+        // TODO this is incorrect, the check for man in the middle needs to be done
+        //   on individual proxy level, not on the `proxyConfiguration` level,
```
szmarczak, line 817: I know this is a little bit off-topic, but how this should be solved?

packages/jsdom-crawler/src/internals/jsdom-crawler.ts
```diff
@@ -0,0 +1,1023 @@
...
+                get() {
+                    if (contentType.type !== APPLICATION_JSON_MIME_TYPE) return null;
+                    const jsonString = body!.toString(contentType.encoding);
+                    return JSON.parse(jsonString);
+                },
+            });
+
+            Object.defineProperty(crawlingContext, 'body', {
+                get() {
+                    return body;
```
szmarczak, line 687: <details>
<summary>There's been a change</summary>

https://github.com/apify/apify-ts/blob/5eb05d0a24e019fae7044d00a4d11bd7ec9dd8d8/packages/cheerio-crawler/src/internals/cheerio-crawler.ts#L699-L701

This is incorrect; ifs are not memory guards and do not guarantee that something will not consume memory. Example:

```js
// node --expose-gc demo.js
const crypto = require('crypto');

const f = (() => {
        const GB = 1024 * 1024 * 1024;
        const body = crypto.randomBytes(GB * 1);

        return () => {
                if (true) {
                        return '';
                }

                // Comment the line below and it drops the buffer
                return body;
        };
})();

gc(); // Force GC to run

setTimeout(() => {
        console.log(f());
}, 10000);
```

As long as the scope where the buffer is referenced exists, the buffer is kept in the memory as well.
To free it, we'd have to directly set it to something else.
Also note that there still exists a way to retrieve the body (so this is another proof):

```js
context.contentType.type = 'application/json';
context.json; // this will error because it will try to JSON.parse(htmlContentHere)
```
</details>

packages/jsdom-crawler/src/internals/jsdom-crawler.ts
```diff
@@ -0,0 +1,1023 @@
...
+     * Sadly, there are some websites which use invalid headers. Those are encoded using the UTF-8 encoding.
+     * If those sites actually use a different encoding, the response will be corrupted. You can use
+     * `suggestResponseEncoding` to fall back to a certain encoding, if you know that your target website uses it.
+     * To force a certain encoding, disregarding the response headers, use {@link CheerioCrawlerOptions.forceResponseEncoding}
+     * ```
+     * // Will fall back to windows-1250 encoding if none found
+     * suggestResponseEncoding: 'windows-1250'
+     * ```
+     */
+    suggestResponseEncoding?: string;
```
szmarczak, line 303: I'm in favor of removing this option completely and doing what Chrome does. Having this as an option just adds complexity.

packages/jsdom-crawler/src/internals/jsdom-crawler.ts
```diff
@@ -0,0 +1,1023 @@
...
+        if (this.useSessionPool) {
+            this.persistCookiesPerSession = persistCookiesPerSession ?? true;
+        } else {
+            this.persistCookiesPerSession = false;
+        }
+    }
+
+    /**
+     * **EXPERIMENTAL**
+     * Function for attaching CrawlerExtensions such as the Unblockers.
```
szmarczak, line 606: I think those should just extend `got-scraping` instead

packages/jsdom-crawler/src/internals/jsdom-crawler.ts
B4nan: we can measure once we merge, for now i would use the same

packages/jsdom-crawler/src/internals/jsdom-crawler.ts
B4nan: as [the issue description](https://github.com/apify/apify-ts/issues/59) says (and it is a single sentence), there should be a common predecessor for cheerio and jsdom crawlers (it could be a new package, or as long as it does not require additional dependencies, we could also add it to the `@crawlee/basic` package). it should be above basic crawler most probably, but i can as well imagine we move some methods down to the basic crawler if it makes sense. keep in mind that basic crawler is a separate unit, not an abstract class, so it needs to stay working on its own (we cant define abstract methods/props and force users to extend it)

we cant reuse cheerio interfaces here, even if they would be the same, we should use type aliases or interface extensions. reusing them would be very confusing for users as well as for us later during maintenance.

regarding webassmebly, i can imagine benefits, but also huge hickups during maintenance, especially when others will need to do changes in the code they havent written. lets optimize when we see it makes sense.

B4nan (commented):
i will review the crawler implementation once we have the common predecessor. right now i would be reading 1000 loc of mostly code duplicates, i dont want to read it, nor i want to maintain it :]

test/apify/crawlers/jsdom_crawler.test.ts
```diff
@@ -0,0 +1,55 @@
...
+                expect(window.document.title).toBe('foobar');
+                expect(window.document.querySelector('p').textContent).toBe('Hello, world!');
+            },
+            failedRequestHandler: ({ error }) => {
+                failPromise = Promise.reject(error);
+            },
+        });
+
+        await crawler.run();
+        await failPromise;
```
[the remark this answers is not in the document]
szmarczak: In AVA this will fail the test:

```js
test('1234', async (t) => {
  try {
    t.fail('lol');
  } catch {}

  t.pass();
});
```

but `expect` doesn't work like that. I just want to forcibly fail a test.

test/apify/crawlers/jsdom_crawler.test.ts
```diff
@@ -0,0 +1,55 @@
+import http from 'http';
+import { AddressInfo } from 'net';
+import { JSDOMCrawler, RequestList } from '../../../packages/jsdom-crawler/src/index';
```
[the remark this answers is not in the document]
szmarczak: Honestly I'd prefer if the tests were stored along with the source code for the crawler instead

packages/jsdom-crawler/src/internals/jsdom-crawler.ts
```diff
@@ -0,0 +1,1023 @@
...
+    systemStatusOptions: {
+        maxEventLoopOverloadedRatio: 0.7,
+    },
+};
+
+export interface CheerioFailedRequestHandlerInput<JSONData = Dictionary> extends CrawlerHandleFailedRequestInput, CheerioRequestHandlerInputs<JSONData> {}
+
+export type CheerioFailedRequestHandler<JSONData = Dictionary> = (inputs: CheerioFailedRequestHandlerInput<JSONData>) => Awaitable<void>;
+
+export interface CheerioCrawlerOptions<JSONData = Dictionary> extends Omit<
```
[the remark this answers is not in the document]
szmarczak: https://github.com/apify/apify-ts/pull/107#discussion_r873275459

PR #108, opened 2022-05-23
feat: add sendRequest to basic crawler context

### LINKED ISSUE #74: Passing Request to got-scraping?
we need to do something about this, so we support passing request instances directly. either at the SDK level, or in got-scraping. i guess it would be simpler to support `body` as an alias to payload in got-scraping, right?

_Originally posted by @B4nan in https://github.com/apify/apify-ts/pull/62#discussion_r844249161_

szmarczak:
- [x] cookies
- [ ] docs

Closes #74


packages/basic-crawler/src/internals/basic-crawler.ts
```diff
@@ -730,6 +732,20 @@ export class BasicCrawler<
                     requestQueue: await this.getRequestQueue(),
                 });
             },
+            sendRequest: async (options?: OptionsInit) => {
```
szmarczak, line 735: This is quite ugly in the terms of types, because we don't know whether it's a Promise or a stream. So people will need to do:

```ts
const response = await crawlingContext.sendRequest() as Response<string>;
```

packages/basic-crawler/src/internals/basic-crawler.ts
```diff
@@ -730,6 +732,20 @@ export class BasicCrawler<
...
+                return gotScraping({
+                    url: request!.url,
+                    method: request!.method,
+                    body: request!.payload,
+                    headers: request!.headers,
+                    ...options,
```
szmarczak, line 741: What about cookies? There are two possible options.

1. Pass a cookieJar

```ts
async requestHandler({ sendRequest, cookieJar }) {
    const response = await sendRequest({ cookieJar }) as Response<string>;
	...
},
```

2. Pass a cookie header (fire-and-forget)

```ts
async requestHandler({ sendRequest, cookieJar, request }) {
    const response = await sendRequest({
		headers: {
				cookie: cookieJar.getCookieStringSync(request.url),
		},
	}) as Response<string>;
	...
},
```

I'm against the latter because it doesn't work when redirecting back and forth (so this may transform into an anti-pattern). Also currently we cannot access `session.cookieJar` because it's private.

packages/basic-crawler/src/internals/basic-crawler.ts
szmarczak: I feel like we should drop the `isStream` option in Got and support only `got(...)` and `got.stream(...)`.

packages/basic-crawler/src/internals/basic-crawler.ts
```diff
@@ -730,6 +732,20 @@ export class BasicCrawler<
...
+                    retry: {
+                        limit: 0,
+                    },
+                    proxyUrl: crawlingContext.proxyInfo?.url,
+                    sessionToken: session,
```
szmarczak, line 746: Those are required and must not be overwritten.

packages/basic-crawler/src/internals/basic-crawler.ts
B4nan: If we have the `isStream` toggle inside `OptionsInit`, we can easily alter the return type based on that. I can help with that if you don't know how - ideally prepare some examples/tests where you use casting so I have a ground for testing.

Now seeing you have the `options` defined as optional, that's obviously not gonna work for such case, but I would expect users to pass the requests explicitly. And more importantly - the idea behind this was not to have a shotcut for `gotScraping` to fire the current request object, **but to allow processing any `Request` objects** (makes sense to process the bound request by default, but if provided, the request from arguments needs to be used on its own, not as an extension (= those should not be merged to a single options object). The `Request` interface is not compatible with `OptionsInit` interface, right? That was the whole point for this helper method.

> I feel like we should drop the isStream option in Got and support only got(...) and got.stream(...).

Yeah that makes a lot of sense to me, but let's focus on what we can do now.

szmarczak: I see now. Will redo.

B4nan:
Damn turbo cache, its so broken :/ Will merge it, should be working fine apart from the CI failures caused by turbo.

Made some small changes myself, namely making the request param optional, falling back to the contextual request.

PR #115, opened 2022-05-27
test: enable storage-local for e2e tests

AndreyBykov:
By default `process.env.STORAGE_IMPLEMENTATION` is set to `MEMORY` now (as it's default for `Actor`). Tried to run it in CI - finished successfully.

For the `PLATFORM` - it would just run on the platform as before (platform tests are currently running, but it should be fine).

If set to `LOCAL` - `@apify/storage-local` will be used - tried locally - finished with no issues.

UPD: platform tests expectedly finished successfully.

### DIFF
--- a/test/e2e/autoscaling-max-tasks-per-minute/actor/main.js
+++ b/test/e2e/autoscaling-max-tasks-per-minute/actor/main.js
@@ -1,11 +1,17 @@
 import { Actor } from 'apify';
 import { BasicCrawler, log as defaultLog, LogLevel } from '@crawlee/basic';
+import { ApifyStorageLocal } from '@apify/storage-local';
 
 const crawlerLogger = defaultLog.child({
     prefix: 'AutoscalingTest',
     // Set this to false if you want to see verbose output from the autoscaled pool
     level: true ? defaultLog.getOptions().level : LogLevel.PERF,
 });
+
+const mainOptions = {
+    exit: Actor.isAtHome(),
+    storage: process.env.STORAGE_IMPLEMENTATION === 'LOCAL' ? new ApifyStorageLocal() : undefined,
+};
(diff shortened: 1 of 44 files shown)


test/e2e/autoscaling-max-tasks-per-minute/actor/main.js
```diff
@@ -1,12 +1,18 @@
...
 import { BasicCrawler, log as defaultLog, LogLevel } from '@crawlee/basic';
+import { ApifyStorageLocal } from '@apify/storage-local';
 
 const crawlerLogger = defaultLog.child({
     prefix: 'AutoscalingTest',
     // Set this to false if you want to see verbose output from the autoscaled pool
     level: true ? defaultLog.getOptions().level : LogLevel.PERF,
 });
 
+const mainOptions = {
```
vladfrangu, line 11: Should we maybe make this a common util function instead of duplicating the object definition? Can we even do that?

vladfrangu (approved):
LGTM, just one small thought/suggestion/question

test/e2e/autoscaling-max-tasks-per-minute/actor/main.js
AndreyBykov: Probably not, as this `actor` folder is pushed to the platform as-is, thus we don't push files out of it to the platform -> it will just fail there as it would not know from where it should import this function...

vladfrangu: Ah, right, forgot about that, sorry! Still, looks fine to me 👍

B4nan (approved):
I am quite sure we can simplify/deduplicate many things, but let's first aim to have things working :]

PR #122, opened 2022-05-31
test: optimise some e2e tests

AndreyBykov:
- Lowered max number of requests for pagination tests.
- provided specific selector for enqueueLinks tests + limited number of requests there.
Each test out of what we have now should finish with ~1m tops locally.

Will update memory for platform tests + would wait before at least one full e2e run will finish before merging.


B4nan (approved):

AndreyBykov:
4GB memory platform run took 40+ minutes
8GB: ~35 minutes
16GB: ~25 minutes, another run took 50+ again though.

However - merging 16GB. We could always change it later (probably the platform was overloaded, let's see after few more runs).

PR #124, opened 2022-06-02
docs: update docker images + proxy management guides

AndreyBykov:
@B4nan could you please a quick look?

In the docker images - I felt like we should mention the multi-stage build for TS projects.

Proxy management - there was a mention of Apify proxy already, and I am not sure how we should proceed. Now I basically added more info on two options (using Apify proxy vs using own proxies). On one hand - it's about proxy and it should all be there. On the other hand - if we want to separate Apify and Crawlee - maybe it qualifies for "more verbose platfrom guide". But on the other hand - it's not really about the platform, it could also be used locally, but requires the usage of Actor class (and thus apify package). Maybe there shold even be another guide - about using the Actor class locally (proxy, storages etc). What do you think?

### DIFF
--- a/docs/guides/avoid_blocking.mdx
+++ b/docs/guides/avoid_blocking.mdx
@@ -8,18 +8,23 @@ Browser fingerprint is a collection of browser attributes and significant featur
 
 Changing browser fingerprints can be a tedious job. Luckily, Crawlee provides this feature out of the box with zero configuration necessary. Let's take a look at how it is done.
 
- Changing browser fingerprints is available in `PuppeteerCrawler` and `PlaywrightCrawler`. You have to pass the `useFingerprints` option to the `browserPoolOptions`.
+ Changing browser fingerprints is available in `PlaywrightCrawler` and `PuppeteerCrawler`. You have to pass the `useFingerprints` option to the `browserPoolOptions`.
 
  ```javascript
+import { PlaywrightCrawler } from '@crawlee/playwright';
+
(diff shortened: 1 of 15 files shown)


B4nan:
> But on the other hand - it's not really about the platform, it could also be used locally, but requires the usage of Actor class (and thus apify package)

It requires the platform, apify proxy = platform. Yes, you can use apify proxy locally, just like you can with apify API/storage. Crawlee docs should mention only the `ProxyConfiguration` implementation from core package + I guess we can link the apify platform guide from there once we have the information there. The difference is something to describe in the platform section, not in crawlee related docs.

I would suggest using import statements in the examples so its clear where is that class coming from. Because the apify implementation needs to be created via the async factory, not constructor.

AndreyBykov:
Could you guys have a look?

@vladfrangu with the idea of having the snippets in separate files - are you aware of some way to calm down TS/eslint? Since I don't have fully working examples there - there always be missing some param, etc. Or are we fine to have almost every TS file there highlighted?

@B4nan I startet to add these tabs with codeblocks, some were there, some I added. Do we want to add it everywhere? E.g. in avoid blocking there's only Playwright snippeto to `useFingerprints`. I guess it would be nice to be consistent and have it for all crawlers (same as it's currently in Proxy management guide)?

B4nan:
It's indeed nice to have but at this stage I'd rather focus on content than appearence.

AndreyBykov:
@B4nan So let's merge this one and I'll continue further?

PR #125, opened 2022-06-02
docs: cleanup examples and make them more accurate

vladfrangu:
This is part one which mostly focuses on enqueueLink as well as cleaning up examples/guides

A future PR will come for Request.skipNavigation

### DIFF
--- a/docs/examples/basic_crawler.ts
+++ b/docs/examples/basic_crawler.ts
@@ -12,7 +12,9 @@ const crawler = new BasicCrawler({
         console.log(`Processing ${url}...`);
 
         // Fetch the page HTML via the crawlee sendRequest utility method
-        const { body } = await sendRequest(request);
+        // By default, the method will use the current request that is being handled, so you don't have to
+        // provide it yourself. You can also provide a custom request if you want.
+        const { body } = await sendRequest();
 
         // Store the HTML and URL to the default dataset.
         await dataset.pushData({
--- a/docs/examples/basic_crawler.ts
+++ b/docs/examples/basic_crawler.ts
@@ -1,5 +1,4 @@
 import { BasicCrawler, Dataset } from '@crawlee/basic';
-import { gotScraping } from 'got-scraping';
 
 // Create a dataset where we will store the results.
 const dataset = await Dataset.open();
@@ -8,16 +7,16 @@ const dataset = await Dataset.open();
 // users to implement the crawling logic themselves.
 const crawler = new BasicCrawler({
     // This function will be called for each URL to crawl.
-    async requestHandler({ request }) {
+    async requestHandler({ request, sendRequest }) {
         const { url } = request;
         console.log(`Processing ${url}...`);
 
-        // Fetch the page HTML via Apify utils gotScraping
-        const { body } = await gotScraping({ url });
+        // Fetch the page HTML via the crawlee sendRequest utility method
+        const { body } = await sendRequest(request);
 
         // Store the HTML and URL to the default dataset.
         await dataset.pushData({
-            url: request.url,
+            url,
             html: body,
         });
     },
--- a/docs/examples/cheerio_crawler.mdx
+++ b/docs/examples/cheerio_crawler.mdx
@@ -7,9 +7,7 @@ import CodeBlock from '@theme/CodeBlock';
 import ApiLink from '@site/src/components/ApiLink';
 import CheerioCrawlerSource from '!!raw-loader!./cheerio_crawler.ts';
 
-This example demonstrates how to use <ApiLink to="cheerio-crawler/class/CheerioCrawler">`CheerioCrawler`</ApiLink> to crawl a list of URLs from an external
-file, load each URL using a plain HTTP request, parse the HTML using the [Cheerio library](https://www.npmjs.com/package/cheerio) and extract some data from it:
-the page title and all `h1` tags.
+This example demonstrates how to use <ApiLink to="cheerio-crawler/class/CheerioCrawler">`CheerioCrawler`</ApiLink> to crawl a list of URLs from an external file, load each URL using a plain HTTP request, parse the HTML using the [Cheerio library](https://www.npmjs.com/package/cheerio) and extract some data from it: the page title and all `h1` tags.
 
 <CodeBlock className="language-js">
 	{CheerioCrawlerSource}
--- a/docs/examples/cheerio_crawler.mdx
+++ b/docs/examples/cheerio_crawler.mdx
@@ -7,7 +7,9 @@ import CodeBlock from '@theme/CodeBlock';
 import ApiLink from '@site/src/components/ApiLink';
 import CheerioCrawlerSource from '!!raw-loader!./cheerio_crawler.ts';
 
-This example demonstrates how to use <ApiLink to="cheerio-crawler/class/CheerioCrawler">`CheerioCrawler`</ApiLink> to crawl a list of URLs from an external file, load each URL using a plain HTTP request, parse the HTML using the [Cheerio library](https://www.npmjs.com/package/cheerio) and extract some data from it: the page title and all `h1` tags.
+This example demonstrates how to use <ApiLink to="cheerio-crawler/class/CheerioCrawler">`CheerioCrawler`</ApiLink> to crawl a list of URLs from an external
+file, load each URL using a plain HTTP request, parse the HTML using the [Cheerio library](https://www.npmjs.com/package/cheerio) and extract some data from it:
+the page title and all `h1` tags.
 
 <CodeBlock className="language-js">
 	{CheerioCrawlerSource}
--- a/docs/examples/crawl_some_links.ts
+++ b/docs/examples/crawl_some_links.ts
@@ -9,7 +9,7 @@ const crawler = new CheerioCrawler({
         console.log(request.url);
         // Add some links from page to the crawler's RequestQueue
         await enqueueLinks({
-            pseudoUrls: ['http[s?]://apify.com/[.+]/[.+]'],
+            globs: ['apify.com/*/*'],
         });
     },
 });
--- a/docs/examples/crawl_some_links.ts
+++ b/docs/examples/crawl_some_links.ts
@@ -9,7 +9,7 @@ const crawler = new CheerioCrawler({
         console.log(request.url);
         // Add some links from page to the crawler's RequestQueue
         await enqueueLinks({
-            globs: ['apify.com/*/*'],
+            globs: ['https://apify.com/*/*'],
         });
     },
 });
--- a/docs/examples/crawl_some_links.ts
+++ b/docs/examples/crawl_some_links.ts
@@ -9,7 +9,7 @@ const crawler = new CheerioCrawler({
         console.log(request.url);
         // Add some links from page to the crawler's RequestQueue
         await enqueueLinks({
-            globs: ['https://apify.com/*/*'],
+            globs: ['http?(s)://apify.com/*/*'],
         });
     },
 });
--- a/docs/examples/map.ts
+++ b/docs/examples/map.ts
@@ -1,6 +1,6 @@
 import { Dataset, KeyValueStore } from '@crawlee/core';
 
-const dataset = await Dataset.open<{ headingCount: number }>();
+const dataset = await Dataset.open();
 const keyValueStore = await KeyValueStore.open();
 
 // calling map function and filtering through mapped items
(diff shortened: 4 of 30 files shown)


docs/examples/crawl_some_links.ts
```diff
@@ -9,7 +9,7 @@ const crawler = new CheerioCrawler({
         console.log(request.url);
         // Add some links from page to the crawler's RequestQueue
         await enqueueLinks({
-            pseudoUrls: ['http[s?]://apify.com/[.+]/[.+]'],
+            globs: ['apify.com/*/*'],
```
vladfrangu, line 12: @AndreyBykov is this the right way to convert this? It looks fine to me but I don't know if I need to use two `*` or not

AndreyBykov: I think it should be something like `?(http|https)://apify.com/**/*` - but I am not like a huge expert in globs, so would double check it still :/

But in pseudoUrls tests `[http|https]://cool.com/` became `?(http|https)://cool.com/` and `https://example.com/[(\w|-|/)*]` -> `https://example.com/**/*` (and it was verified that it still works the same :) )

B4nan: Maybe we could strip the protocol before we do the matching?

docs/examples/basic_crawler.ts
```diff
@@ -8,16 +7,16 @@ const dataset = await Dataset.open();
...
     // This function will be called for each URL to crawl.
-    async requestHandler({ request }) {
+    async requestHandler({ request, sendRequest }) {
         const { url } = request;
         console.log(`Processing ${url}...`);
 
-        // Fetch the page HTML via Apify utils gotScraping
-        const { body } = await gotScraping({ url });
+        // Fetch the page HTML via the crawlee sendRequest utility method
+        const { body } = await sendRequest(request);
```
B4nan, line 15: worth mentioning that `sendRequest()` will do the same here, as it takes the contextual `request` object as default value

docs/examples/cheerio_crawler.mdx
```diff
@@ -7,7 +7,9 @@ import CodeBlock from '@theme/CodeBlock';
 import ApiLink from '@site/src/components/ApiLink';
 import CheerioCrawlerSource from '!!raw-loader!./cheerio_crawler.ts';
 
-This example demonstrates how to use <ApiLink to="cheerio-crawler/class/CheerioCrawler">`CheerioCrawler`</ApiLink> to crawl a list of URLs from an external file, load each URL using a plain HTTP request, parse the HTML using the [Cheerio library](https://www.npmjs.com/package/cheerio) and extract some data from it: the page title and all `h1` tags.
+This example demonstrates how to use <ApiLink to="cheerio-crawler/class/CheerioCrawler">`CheerioCrawler`</ApiLink> to crawl a list of URLs from an external
+file, load each URL using a plain HTTP request, parse the HTML using the [Cheerio library](https://www.npmjs.com/package/cheerio) and extract some data from it:
```
B4nan, line 11: not sure if those line breaks are coming from some auto formatting. i would rather not see line breaks in text, line wraps should be handled by IDE

docs/examples/map.ts
```diff
@@ -1,6 +1,6 @@
 import { Dataset, KeyValueStore } from '@crawlee/core';
 
-const dataset = await Dataset.open<{ headingCount: number }>();
+const dataset = await Dataset.open();
```
B4nan, line 3: why? we want to show generics usage

docs/examples/puppeteer_crawler_crawler_utils_snapshot.ts
```diff
@@ -1,12 +1,12 @@
-import { PuppeteerCrawler, puppeteerUtils } from '@crawlee/puppeteer';
+import { PuppeteerCrawler } from '@crawlee/puppeteer';
 
 // Create a PuppeteerCrawler
 const crawler = new PuppeteerCrawler({
-    async requestHandler({ request, page }) {
+    async requestHandler({ request, saveSnapshot }) {
         // Convert the URL into a valid key
         const key = request.url.replace(/[:/]/g, '_');
```
B4nan, line 7: i guess we could make this work automatically and make the `key` param optional

docs/examples/puppeteer_recursive_crawl.ts
```diff
@@ -6,7 +6,7 @@ const crawler = new PuppeteerCrawler({
         console.log(`Title of ${request.url}: ${title}`);
 
         await enqueueLinks({
-            pseudoUrls: ['https://www.iana.org/[.*]'],
+            globs: ['https://www.iana.org/*'],
```
B4nan, line 9: i dont think this is the same, single `*` matches up until the first `/`, you need `**`, and maybe even without the initial slash (please test this, i am not 100% sure)

docs/guides/getting_started.mdx
```diff
@@ -130,27 +130,41 @@ introduce all Crawlee classes necessary to make it happen.
 
 ### The general idea
 
-There are 4 crawler classes available for use in Crawlee. <ApiLink to="basic-crawler/class/BasicCrawler">`BasicCrawler`</ApiLink>, <ApiLink to="cheerio-crawler/class/CheerioCrawler">`CheerioCrawler`</ApiLink>, <ApiLink to="puppeteer-crawler/class/PuppeteerCrawler">`PuppeteerCrawler`</ApiLink> and <ApiLink to="playwright-crawler/class/PlaywrightCrawler">`PlaywrightCrawler`</ApiLink>. We'll talk about their differences later. Now, let's talk about what they have in common.
+There are 4 crawler classes available for use in Crawlee. <ApiLink to="basic-crawler/class/BasicCrawler">`BasicCrawler`</ApiLink>, <ApiLink to="cheerio-crawler/class/CheerioCrawler">`CheerioCrawler`</ApiLink>, <ApiLink to="puppeteer-crawler/class/PuppeteerCrawler">`PuppeteerCrawler`</ApiLink> and <ApiLink to="playwright-crawler/class/PlaywrightCrawler">`PlaywrightCrawler`</ApiLink>.
```
B4nan, line 133: those new line breaks are making it quite hard to see what actually changed :/ please remove them and make sure this wont happen again

or is it coming from the prettier in docs? just realized we have it there. tbh i was never a fan of this tool, as it produces quite ugly code when it comes to line breaks (talking about actual code, not docs)...

website/src/pages/index.js
```diff
@@ -81,7 +89,7 @@ function ActorExample() {
     return (
         <section id="try" className="container">
             <h2>Try it out</h2>
-            <p>Install Apify SDK into a Node.js project. You must have Node.js 16 or higher installed.</p>
+            <p>Install the Crawlee SDK into a Node.js project. You must have Node.js 16 or higher installed.</p>
```
B4nan, line 92:
```suggestion
            <p>Install Crawlee into a Node.js project. You must have Node.js 16 or higher installed.</p>
```

docs/examples/cheerio_crawler.mdx
vladfrangu: The ones I did were manually done actually. I can revert them, but then we'll have inconsistencies between docs in different or even the same files, as some have line breaks, and some don't... We should probably really consider setting up a tool to format docs consistently so we don't have issues in the future

docs/examples/map.ts
B4nan: The file is called `map.ts`. i want all examples to be TS.

docs/examples/puppeteer_crawler_crawler_utils_snapshot.ts
vladfrangu: Maybe instead this example should use the page title and not the url? Or just omit key entirely as it is not required

docs/examples/puppeteer_recursive_crawl.ts
vladfrangu: Will test to make sure it still works, and correct the patterns otherwise

docs/examples/cheerio_crawler.mdx
B4nan: We already have inconsistencies. We should gradually fix them to the right state = remove line breaks, not add them :]

Not sure if you can force pretties to remove line breaks, I dont think so, its very "inconfigurable" (another thing why I dislike it).

docs/examples/puppeteer_crawler_crawler_utils_snapshot.ts
B4nan: Oh so its not required already, I guess you can ignore me then :] It was more of an idea for implementation than comment for the example.

docs/examples/puppeteer_recursive_crawl.ts
B4nan: When you are there, also check whether `https?://www.iana.org/` will match both https and http, I believe it should. Some examples where allowing both protocols, and we should show how to do that with globs too.

docs/examples/cheerio_crawler.mdx
vladfrangu: Hah, I used to feel the same way about it, then I started using it and I couldn't go back from it (in projects that use it at least) 😄 (it's especially useful when it splits up long lines in a decently readable way, at least if you configure the max width to _not_ be 80 but make it a more reasonable 120-150).

That said, I will revert the line breaks. I still think we should keep them, as it will make seeing diffs easier, but I'll make a PR after this one to bring the rest of them to no line breaks

B4nan: I used to maintain line breaks in docs manually too. In retrospective, it was a very bad decision, the editor width is never the same :] 

Other than the line breaking issue, I would be quite ok with prettier, the one last thing I kinda hate about it is the funky removal of blank lines that is totally inconfigurable (let me just disable that!). They even say "its hard problem so we keep things as they are" or something like that, and then they remove the blank lines inside class definitions etc :D 

I just can't adhere to code style decisions of some "prettier god who thinks knows best". Make things configurable for christ sake :D

docs/examples/puppeteer_recursive_crawl.ts
vladfrangu: So:

1. To have the same matching, it is with double `*`
2. `http?://www.iana.org/**` will match http, https and probably any variant of `http<letter>`. If that's fine I'll add it in the example that had optional https

AndreyBykov: `http?` will not match `http` though (just checked)

AndreyBykov: looks like `http?(s)://www.iana.org/**` is what we need here

docs/examples/crawl_some_links.ts
vladfrangu: Just tried the globs, the most accurate one is `http?://apify.com/*/**`, although if you want be extra strict (so it only matches apify.com/<username>/<actor>`, one `*` per `/` is ideal

docs/examples/puppeteer_recursive_crawl.ts
AndreyBykov: Check [here](https://www.digitalocean.com/community/tools/glob?comments=false&glob=http%3F%28s%29%3A%2F%2Fwww.iana.org%2F%2A%2A&matches=false&tests=http%3A%2F%2Fwww.iana.org&tests=https%3A%2F%2Fwww.iana.org&tests=httpa%3A%2F%2Fwww.iana.org&tests=http%3A%2F%2Fwww.iana.org%2F&tests=https%3A%2F%2Fwww.iana.org%2F&tests=httpa%3A%2F%2Fwww.iana.org%2F&tests=http%3A%2F%2Fwww.iana.org%2Fsub&tests=https%3A%2F%2Fwww.iana.org%2Fsub&tests=httpa%3A%2F%2Fwww.iana.org%2Fsub&tests=http%3A%2F%2Fwww.iana.org%2Fsub%2Fdomain&tests=https%3A%2F%2Fwww.iana.org%2Fsub%2Fdomain&tests=httpa%3A%2F%2Fwww.iana.org%2Fsub%2Fdomain)

vladfrangu: Will that match http *and* https? If so..that is some confusing syntax globs have 😅

vladfrangu: Sweet, thanks! Will change

AndreyBykov: It is confusing indeed :)

B4nan: oh god, it means "one character" and not zero or one, wth

B4nan: maybe we should support `http?://` prefix automatically (rewrite it to the right form before we pass it to minimatch) :]

B4nan (approved):

PR #129, opened 2022-06-06
test: add platform links to e2e tests

AndreyBykov:
PLATFORM e2e tests take significantly more time than local/memory storage ones mainly due to actor build, now we'll have a link to build log with build time + view run link with actual run time (without build time).

CI run [here](https://github.com/apify/apify-ts/runs/6758395569?check_suite_focus=true).

I was also annoyed by missing spaces in logs - it somehow ignores the space after the `colors[color]` is used it seems, so moved the space inside `colors[color]` calls - kinda shitty solution, but at least I am not triggered every time now =D


B4nan:
I also spotted those missing spaces today :D 

Good, btw those longer builds times are really caused by the new fingerprint generator and its dependencies, @barjin is already working on fix for that.

AndreyBykov:
nice :) would be great to further lower down the run time :)

PR #134, opened 2022-06-09
docs: update example in puppeteerUtils

mvolfik:
I also migrated it to ESM syntax, as that seems to be what everything is moving towards (especially templates)


mvolfik:
Why my PRs with docs changes always functional tests :anger:

B4nan (approved):

PR #135, opened 2022-06-14
refactor: move caching of storage managers to the `StorageManager` class

B4nan:
This way the static helpers on storage classes can implement caching themselves.


packages/core/src/storages/key_value_store.ts
```diff
@@ -141,16 +144,30 @@ export class KeyValueStore {
...
+    }
+
+    async getAutoSavedValue<T extends Dictionary = Dictionary>(key: string, defaultValue = {} as T): Promise<T> {
+        if (this.cache.has(key)) {
+            return this.cache.get(key) as T;
+        }
+
+        const value = await this.getValue<T>(key, defaultValue);
+        this.cache.set(key, value);
+        this.config.getEventManager().on('persistState', () => this.setValue(key, value));
```
vladfrangu, line 168: Do we want to add a listener every time someone sets a value? Can we store references instead and call the listener once?

packages/core/src/storages/key_value_store.ts
```diff
@@ -213,6 +230,23 @@ export class KeyValueStore {
         // Make copy of options, don't update what user passed.
         const optionsCopy = { ...options };
 
+        // If we try to set the value of a cached state to a different reference, we need to update the cache accordingly.
+        const cachedValue = this.cache.get(key) as T;
+
+        if (key in this.cache && cachedValue !== value) {
```
vladfrangu, line 236: Did you mean this.cache.has(key)?

B4nan: good catch, i had a POJO there initially

packages/core/src/storages/key_value_store.ts
B4nan: wdym? we only add it once, the value is cached, so this branch will be executed once per key only (for a given instance of KV store)

the use case here is to have persistent value that will be always saved on the persist state event, which is (among other cases) fired during actor migration on platform

B4nan: > especially if someone calls that method with the same key multiple times

That's why we cache the value, as I said, this branch will be executed only once and then the cached reference will be used instead for following calls.

Usually there will be just one value here (so one listener), the crawler state.

vladfrangu: Woops, yeah sorry I was expressing myself wrong (meant multiple different keys), but it should be fine nevertheless. Just a thought

PR #136, opened 2022-06-15
Docs/guides 02

AndreyBykov:
* Quick Start (added CLI)
* Result Storage

fixed broken link in crawl_some_links example


AndreyBykov:
would appreciate a quick look at result storage (whether it's enough etc)

docs/guides/quick_start_cheerio.ts
```diff
@@ -0,0 +1,20 @@
...
+        console.log(`Title of ${url}: ${title}`);
+        // Add URLs that match the provided pattern.
+        await enqueueLinks({
+            globs: ['https://www.iana.org/*'],
+        });
+        // Save extracted data to dataset.
+        await Dataset.pushData({ url, title });
+    },
+});
+// Choose the first URL to open.
```
vladfrangu, line 17: Really a nit but maybe some new lines would be nice?
```suggestion
});

// Choose the first URL to open.
```

docs/guides/result_storage.mdx
```diff
@@ -5,106 +5,96 @@ title: Result Storage
...
-{APIFY_LOCAL_STORAGE_DIR}/key_value_stores/{STORE_ID}/{KEY}.{EXT}
+{CRAWLEE_STORAGE_DIR}/key_value_stores/{STORE_ID}/{KEY}.{EXT}
 ```
 
-Note that `{STORE_ID}` is the name or ID of the key-value store. The default key value store has ID `default`, unless you override it by setting the
-`APIFY_DEFAULT_KEY_VALUE_STORE_ID` environment variable. The `{KEY}` is the key of the record and `{EXT}` corresponds to the MIME content type of the
-data value.
+:::note
+
+`{STORE_ID}` is the name or the ID of the key-value store. The default key-value store has ID `default`, unless you override it by setting the `APIFY_DEFAULT_KEY_VALUE_STORE_ID` environment variable. The `{KEY}` is the key of the record and `{EXT}` corresponds to the MIME content type of the data value.
```
vladfrangu, line 28: This remark about Apify might be better in the platform section?

AndreyBykov: Well - we discussed that crawlee guides should not mention Apify in any way, that there is platform guide, which would be more verbose, and all Apify-related stuff would be there. In this case - there should probably be some other ENV var that could be used with Crawlee locally to override the default storage name. So it's not missed - it should be replaced...

docs/guides/result_storage.mdx
```diff
@@ -5,106 +5,96 @@ title: Result Storage
...
 
 The following code demonstrates basic operations of key-value stores:
 
 ```javascript
-// Get actor input from the default key-value store
-const input = await Actor.getInput();
+import { KeyValueStore } from '@crawlee/core';
+//                   ... from '@crawlee/cheerio';
+//                   ... from '@crawlee/playwright';
+//                   ... from '@crawlee/puppeteer';
```
B4nan, line 38: just FYI we will be using the `crawlee` metapackage in docs to simplify things, but you can keep it here, we can do a quick find/replace and fix it in the whole docs in another PR

docs/guides/result_storage.mdx
B4nan: yop, we will need most of the env vars to work with `CRAWLEE_` prefix, no need to care about it now as they dont exist yet, but thats something to fix in following week or two

docs/guides/result_storage.mdx
```diff
@@ -114,3 +104,18 @@ await dataset.pushData([{ foo: 'bar2', col2: 'val2' }, { col3: 123 }]);
...
+
+Default storages are not purged before the crawler starts if not specified otherwise. If you want to start the run with the empty default storages, you should first <ApiLink to="core/class/Configuration#getStorageClient">get</ApiLink> the default <ApiLink to="types/interface/StorageClient">`StorageClient`</ApiLink> instance and then use the <ApiLink to="types/interface/StorageClient#purge">`purge()`</ApiLink> function before running the crawler as follows:
+
+```javascript
+import { Configuration } from '@crawlee/core';
+//                   ... from '@crawlee/cheerio';
+//                   ... from '@crawlee/playwright';
+//                   ... from '@crawlee/puppeteer';
+
+await Configuration.getStorageClient().purge();
```
B4nan, line 118: we should probably add a `purgeStorage` function that will do this so users dont have to write such a weird line :]

B4nan (approved):
just some notes so we are more aligned, will merge this right ahead so i can see how it looks when rendered :]

PR #137, opened 2022-06-15
feat: session-based fingerprints

barjin:
as mentioned in apify/browser-pool#64

Allows browser fingerprints to be based on Session (instead of current proxy URL).


B4nan (approved):
otherwise it looks good to me. maybe one thing to double check, afaik session pool is optional (enabled by default), so we might wanna fallback to the proxyUrl as the key? or just double check how it works with session pool disabled

PR #139, opened 2022-06-15
chore: fix flaky memory storage test

vladfrangu:
Fixes https://github.com/apify/apify-ts/pull/138#issuecomment-1156672074

Since the worker can take a while to actually write the files, a set timeout might not be sufficient (especially if overloaded). Now the test will keep trying to access the dir (since we know it should be created) before running tests on it


B4nan (approved):

vladfrangu:
Merge ready, the CI errored with something I've never seen before

PR #143, opened 2022-06-21
feat: purge default storages even without `Actor.init`

B4nan:


AndreyBykov (approved):

CHANGELOG.md
```diff
@@ -20,7 +20,7 @@ See [Conventional Commits](https://conventionalcommits.org) for commit guideline
 * (internal) `Request.handledAt` is now string date in ISO format
 * (perf/internal) `Request.inProgress` and `Request.reclaimed` are now `Set`s instead of dictionaries
 * `injectUnderscore` from puppeteer utils has been removed
-* `APIFY_MEMORY_MBYTES` is no longer taken into account, use `AVAILABLE_MEMORY_RATIO` instead
+* `APIFY_MEMORY_MBYTES` is no longer taken into account, use `CRAWLEE_AVAILABLE_MEMORY_RATIO` instead
```
mnmkng, line 23: Ratio? So it's no longer mbytes but percentage of system memory?

packages/apify/src/actor.ts
```diff
@@ -399,7 +396,7 @@ export class Actor<Data extends Dictionary = Dictionary> {
             customAfterSleepMillis = this.config.get('metamorphAfterSleepMillis'),
             ...metamorphOpts
         } = options;
-        const runId = this.config.get('actorRunId')!;
+        const runId = process.env[ENV_VARS.ACTOR_RUN_ID]!;
```
mnmkng, line 399: Why are we coming back to reading process envs? I thought the configuration would read the envs and then everything else would read it from configuration.

CHANGELOG.md
B4nan: Yes, we use 1/4 of available memory by default, and this limit is removed when using `Actor.init` and we are on platform, its based on the [previous implementation](https://github.com/apify/apify-js/blob/5d372cbeade2ec78658d84ea6a3e9893598229b5/src/autoscaling/snapshotter.js#L479). 

But I do plan to support exact setting of memory limit too, to keep the

packages/apify/src/actor.ts
B4nan: Removed those changes for now, I guess we should have that second config class in actor sdk if we want to support those too.

PR #144, opened 2022-06-21
chore: resolve some todos

B4nan:


packages/apify/src/key_value_store.ts
```diff
@@ -0,0 +1,17 @@
...
+     * Returns a URL for the given key that may be used to publicly
+     * access the value in the remote key-value store.
+     */
+    getPublicUrl(key: string): string {
+        return `https://api.apify.com/v2/key-value-stores/${this.id}/records/${key}`;
+    }
+}
+
+// @ts-expect-error
+CoreKeyValueStore.prototype.getPublicUrl = KeyValueStore.prototype.getPublicUrl;
```
vladfrangu, line 17: Can you also module augment @crawlee/core for the kvs?

B4nan: like `declare module ...`? how would that actually work? is that really enough for having that method available even if we keep importing from core package? that would be best

B4nan: This never works well for me, what am I doing wrong? :D It looks exactly as [in the docs](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation)

```ts
import { KeyValueStore } from '@crawlee/core';

declare module '@crawlee/core' {
    interface KeyValueStore {
        /**
         * Returns a URL for the given key that may be used to publicly
         * access the value in the remote key-value store.
         */
        getPublicUrl(key: string): string;
    }
}

// TS2693: 'KeyValueStore' only refers to a type, but is being used as a value here.
KeyValueStore.prototype.getPublicUrl = function getPublicUrl(this: KeyValueStore, key: string): string {
    return `https://api.apify.com/v2/key-value-stores/${this.id}/records/${key}`;
};
```

B4nan: I'd appreciate if you could take a look at this, worst case we keep the correct type only on the class exported from actor sdk.

PR #145, opened 2022-06-22
chore: bundle d.ts files

vladfrangu:
Allows easy module augmentation, as well as keeping everything in one file

Downside: You'll need to build the project once before the editor infers the types for the `@crawlee` modules... It's either this small issue (which is automatically done on npm installs) or having to ts-ignore a lot in apify/key_value_store


vladfrangu:
Anyways:

- I reverted the postinstall hooks as well as the path changes
- The module augmentation is now added as a postbuild step in apify package
- The augmentation happens with a ts-expect-error since there's no way to have both without TS  breaking down
- All declaration files are now bundled into one big d.ts file, which works the same + allows easier augmentation in the future
- Fixed puppeteer's types, since that d.ts did not work fully, and module augmentation is better, but I'd love to test this first somehow (npm pack doesn't seem to do what is expected..guess an @ next would be needed?P

B4nan:
I dont care about E2E/CI tests, I care about local development. If this requires building the project to be able to run tests, I dont want this. Postinstall won't cover anything, we are talking about development, code changes, install happens only once at the very beginning.

Keep in mind we are dealing with very niche feature that can (and imo should) be handled differently. I am not willing to sacriface DX for this.

vladfrangu:
>  If this requires building the project to be able to run tests

Not anymore! I guess I forgot to clarify that in the first bullet point in the comment I wrote today 😓

B4nan:
Yeah I was replying to the first message and saw the second one afterwards, will review when I have time, but it sounds as a good enough solution :] Still what I said about the niche feature applies, I dont want to sacrifice anything for it, we could even have a regular helper function for this, no need to have it on the KVS if you ask me (I dont think it should be there, but apparently it is being used a lot).

PR #151, opened 2022-06-29
feat: add error handler for requests to be retried

AndreyBykov:
I know that I need to add it to Cheerio and other crawlers, but could you have a quick look before I proceed with the rest?

Also thinking - maybe it's worth adding helper `crawler.reclaimRequest()` since with user-provided function for requests to be retried - they would need to either manually do it or just skip it. Or thinking about it now - maybe there should be something like `context.reclainRequest` (which would be `true` by default) and otherwise user would need to set it to false and thus it would be marked as handled)?

### DIFF
--- a/packages/basic-crawler/src/internals/basic-crawler.ts
+++ b/packages/basic-crawler/src/internals/basic-crawler.ts
@@ -65,6 +65,8 @@ const SAFE_MIGRATION_WAIT_MILLIS = 20000;
 
 export type RequestHandler<Context extends CrawlingContext = BasicCrawlingContext> = (inputs: Context) => Awaitable<void>;
 
+export type ErrorHandler<Context extends CrawlingContext = BasicCrawlingContext> = (inputs: Context, error: Error) => Awaitable<void>;
+
 export type FailedRequestHandler<Inputs extends FailedRequestContext = BasicFailedRequestContext> = (inputs: Inputs) => Awaitable<void>;
 
 export interface BasicCrawlerOptions<
@@ -152,23 +154,21 @@ export interface BasicCrawlerOptions<
 
     /**
      * A function to handle requests that failed less than `option.maxRequestRetries` times
-     * and would be retried by the crawler by default.
-     * Note that if the function is user-provided, it would not reclaim the request automatically.
+     * and would be retried by the crawler.
      *
-     * The function receives the following object as an argument:
+     * The function receives the following object as the first argument:
      * ```
      * {
      *     request: Request,
-     *     error: Error,
      *     session: Session,
      *     crawler: BasicCrawler,
      * }
      * ```
-     * where the {@link Request} instance corresponds to the request to be reclaimed, and the `Error` instance
+     * where the {@link Request} instance corresponds to the request to be reclaimed.
+     * Second argument is the `Error` instance that
      * represents the last error thrown during processing of the request.
-     *
      */
-    preReclaimRequestHandler?: FailedRequestHandler<ErrorContext>;
+    errorHandler?: ErrorHandler<Context>;
 
     /**
      * A function to handle requests that failed more than `option.maxRequestRetries` times.
@@ -370,7 +370,7 @@ export class BasicCrawler<
 
     protected log: Log;
     protected requestHandler!: RequestHandler<Context>;
-    protected preReclaimRequestHandler?: FailedRequestHandler<ErrorContext>;
+    protected errorHandler?: ErrorHandler<Context>;
     protected failedRequestHandler?: FailedRequestHandler<ErrorContext>;
     protected requestHandlerTimeoutMillis!: number;
     protected internalTimeoutMillis: number;
@@ -395,7 +395,7 @@ export class BasicCrawler<
         requestHandlerTimeoutSecs: ow.optional.number,
         // TODO: remove in a future release
         handleRequestTimeoutSecs: ow.optional.number,
-        preReclaimRequestHandler: ow.optional.function,
+        errorHandler: ow.optional.function,
         failedRequestHandler: ow.optional.function,
         // TODO: remove in a future release
         handleFailedRequestFunction: ow.optional.function,
@@ -442,7 +442,7 @@ export class BasicCrawler<
             handleRequestTimeoutSecs,
             requestHandlerTimeoutSecs,
 
-            preReclaimRequestHandler,
+            errorHandler,
 
             handleFailedRequestFunction,
             failedRequestHandler,
@@ -466,7 +466,7 @@ export class BasicCrawler<
             this.requestHandler = this.router;
         }
 
-        this.preReclaimRequestHandler = preReclaimRequestHandler;
+        this.errorHandler = errorHandler;
 
         this._handlePropertyNameChange({
             newName: 'failedRequestHandler',
@@ -956,15 +956,22 @@ export class BasicCrawler<
             throw error;
         }
 
-        // @ts-expect-error It is assignable, but TS says otherwise...
-        const castedErrorContext = crawlingContext as ErrorContext;
-        castedErrorContext.error = error;
-
         const shouldRetryRequest = !request.noRetry && request.retryCount < this.maxRequestRetries && !(error instanceof NonRetryableError);
         if (shouldRetryRequest) {
             request.retryCount++;
 
-            await this._handlePreReclaimRequestHandler(castedErrorContext, source);
+            if (this.errorHandler) {
+                await this.errorHandler(crawlingContext, error);
+            }
+
+            const { url, retryCount, id } = request;
+            // We don't want to see the stack trace in the logs by default, when we are going to retry the request.
+            // Thus, we print the full stack trace only when CRAWLEE_VERBOSE_LOG environment variable is set to true.
+            this.log.warning(
+                `Reclaiming failed request back to the list or queue. ${!process.env.CRAWLEE_VERBOSE_LOG ? error : error.stack}`,
+                { id, url, retryCount },
+            );
+            await source.reclaimRequest(request);
         } else {
             // If we get here, the request is either not retryable
             // or failed more than retryCount times and will not be retried anymore.
@@ -973,25 +980,11 @@ export class BasicCrawler<
             await source.markRequestHandled(request);
             this.stats.failJob(request.id || request.uniqueKey);
 
-            await this._handleFailedRequestHandler(castedErrorContext); // This function prints an error message.
-        }
-    }
+            // @ts-expect-error It is assignable, but TS says otherwise...
+            const castedErrorContext = crawlingContext as ErrorContext;
+            castedErrorContext.error = error;
 
-    protected async _handlePreReclaimRequestHandler(crawlingContext: ErrorContext, source: RequestList | RequestQueue): Promise<void> {
-        if (this.preReclaimRequestHandler) {
-            await this.preReclaimRequestHandler(crawlingContext);
-        } else {
-            const { id, url, retryCount } = crawlingContext.request;
-            // We don't want to see the stack trace in the logs by default, when we are going to retry the request.
-            // Thus, we print the full stack trace only when CRAWLEE_VERBOSE_LOG environment variable is set to true.
-            this.log.warning(
-                `Reclaiming failed request back to the list or queue. ${!process.env.CRAWLEE_VERBOSE_LOG
-                    ? crawlingContext.error
-                    : crawlingContext.error.stack
-                }`,
-                { id, url, retryCount },
-            );
-            await source.reclaimRequest(crawlingContext.request);
+            await this._handleFailedRequestHandler(castedErrorContext); // This function prints an error message.
         }
     }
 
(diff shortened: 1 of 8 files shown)


B4nan:
Instead of adding "preReclaim" we should retry on the current `failedRequestHandler` and add new `finalFailedRequestHandler` or maybe even `failedRequestHandlerAfterRetries` to make it self documented. At least that's how we agreed on that with Ondra last time. 

Will check the implementation in a minute, but its weird that this would be handled on cheerio crawler level, I though we will only adjust the basic crawler and it will affect all of them.

packages/basic-crawler/src/internals/basic-crawler.ts
```diff
@@ -150,6 +150,26 @@ export interface BasicCrawlerOptions<
      */
     handleRequestTimeoutSecs?: number;
 
+    /**
+     * A function to handle requests that failed less than `option.maxRequestRetries` times
+     * and would be retried by the crawler by default.
+     * Note that if the function is user-provided, it would not reclaim the request automatically.
```
B4nan, line 156: we should reclaim the request, the purpose here is to allow adjusting the request that will be retried, e.g. labeling it, changing some userData, ...

it should be kind of a hook

packages/basic-crawler/src/internals/basic-crawler.ts
```diff
@@ -930,17 +956,15 @@ export class BasicCrawler<
...
-            const { url, retryCount, id } = request;
-            // We don't want to see the stack trace in the logs by default, when we are going to retry the request.
-            // Thus, we print the full stack trace only when CRAWLEE_VERBOSE_LOG environment variable is set to true.
-            this.log.warning(
-                `Reclaiming failed request back to the list or queue. ${!process.env.CRAWLEE_VERBOSE_LOG ? error : error.stack}`,
-                { id, url, retryCount },
-            );
-            await source.reclaimRequest(request);
+
+            await this._handlePreReclaimRequestHandler(castedErrorContext, source);
```
B4nan, line 967: the code here should stay as it was, just execute the fail handler before it (but after incrementing the retry count)

packages/basic-crawler/src/internals/basic-crawler.ts
```diff
@@ -150,6 +150,26 @@ export interface BasicCrawlerOptions<
...
+     *     error: Error,
+     *     session: Session,
+     *     crawler: BasicCrawler,
+     * }
+     * ```
+     * where the {@link Request} instance corresponds to the request to be reclaimed, and the `Error` instance
+     * represents the last error thrown during processing of the request.
+     *
+     */
+    preReclaimRequestHandler?: FailedRequestHandler<ErrorContext>;
```
mnmkng, line 171: The name is not very intuitive. I'd just go with `failedRequestHandler`. Request reclaiming is an implementation detail of the crawler.

packages/basic-crawler/src/internals/basic-crawler.ts
```diff
@@ -930,17 +956,15 @@ export class BasicCrawler<
             throw error;
         }
 
+        // @ts-expect-error It is assignable, but TS says otherwise...
+        const castedErrorContext = crawlingContext as ErrorContext;
+        castedErrorContext.error = error;
```
mnmkng, line 961: See comment of function comments above 😄

packages/basic-crawler/src/internals/basic-crawler.ts
```diff
@@ -949,14 +973,28 @@ export class BasicCrawler<
...
+            // We don't want to see the stack trace in the logs by default, when we are going to retry the request.
+            // Thus, we print the full stack trace only when CRAWLEE_VERBOSE_LOG environment variable is set to true.
+            this.log.warning(
+                `Reclaiming failed request back to the list or queue. ${!process.env.CRAWLEE_VERBOSE_LOG
+                    ? crawlingContext.error
+                    : crawlingContext.error.stack
+                }`,
+                { id, url, retryCount },
+            );
+            await source.reclaimRequest(crawlingContext.request);
```
mnmkng, line 994: Why is the reclaim inside the `else` clause? The handler should just give the user an opportunity do do something and then continue the same way as before.

If the user wants to prevent the request from being reclaimed, they should set `request.noRetry = true` and the retry mechanism of basic crawler should mark the request as handled.

packages/basic-crawler/src/internals/basic-crawler.ts
AndreyBykov: So basically we would add another option to crawler, but it would not be top-level function, but instead would be called instead of the currently existing one, correct?

AndreyBykov: or actually let me re-read it :)

packages/basic-crawler/src/internals/basic-crawler.ts
B4nan: E.g. you could have a pre nav hook that will make something with the request based on the label, e.g. add some cookies or whatnot. Or you could even try to adjust the payload if its PUT/POST...

mnmkng (changes requested):

packages/basic-crawler/src/internals/basic-crawler.ts
mnmkng: Or maybe just `errorHandler`, that's even more descriptive.

packages/basic-crawler/src/internals/basic-crawler.ts
mnmkng: The use-case for the handler is generic, really: "Let the user do whatever with the request, if there's an error". Because currently, there's no way to do this. If there's an error in the process, the request will immediately go into internal code.

So reclaiming or not reclaiming is not the concern. Quite the contrary, because we don't want to complicate things for the user by forcing them to reclaim manually.

```js
errorHandler: () => { console.log('there was an error') }
``` 

is a valid handler.

AndreyBykov: Yup, get it now 👍

packages/basic-crawler/src/internals/basic-crawler.ts
```diff
@@ -150,6 +150,26 @@ export interface BasicCrawlerOptions<
...
+    /**
+     * A function to handle requests that failed less than `option.maxRequestRetries` times
+     * and would be retried by the crawler by default.
+     * Note that if the function is user-provided, it would not reclaim the request automatically.
+     *
+     * The function receives the following object as an argument:
+     * ```
+     * {
+     *     request: Request,
+     *     error: Error,
```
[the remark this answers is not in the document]
AndreyBykov: I assume it's only for `errorHandler`? If yes - then done. Or you meant to make 2 args for `failedRequestHandler` too?

packages/basic-crawler/src/internals/basic-crawler.ts
```diff
@@ -933,6 +959,11 @@ export class BasicCrawler<
         const shouldRetryRequest = !request.noRetry && request.retryCount < this.maxRequestRetries && !(error instanceof NonRetryableError);
         if (shouldRetryRequest) {
             request.retryCount++;
+
+            if (this.errorHandler) {
+                await this.errorHandler(crawlingContext, error);
+            }
+
```
B4nan, line 966:
```suggestion
            await this.errorHandler?.(crawlingContext, error);
```

packages/browser-crawler/src/internals/browser-crawler.ts
```diff
@@ -171,6 +176,29 @@ export interface BrowserCrawlerOptions<
      */
     handlePageFunction?: BrowserCrawlerHandleRequest<Context>;
 
+    /**
+     * A function to handle requests that failed less than `option.maxRequestRetries` times
```
B4nan, line 180: i'd just say its a function for every retry, and we should say explicitly that it allows modifying the request object before it gets retried

packages/browser-crawler/src/internals/browser-crawler.ts
```diff
@@ -171,6 +176,29 @@ export interface BrowserCrawlerOptions<
...
+     *     request: Request,
+     *     response: Response,
+     *     page: Page,
+     *     browserPool: BrowserPool,
+     *     autoscaledPool: AutoscaledPool,
+     *     session: Session,
+     *     browserController: BrowserController,
+     *     proxyInfo: ProxyInfo,
+     * }
+     * ```
```
B4nan, line 195: we can just point to the interface via `{@link}` macro.

test/core/crawlers/basic_crawler.test.ts
```diff
@@ -267,6 +268,49 @@ describe.each(SingleStorageCase)('BasicCrawler - %s', (Emulator) => {
...
+
+        const basicCrawler = new BasicCrawler({
+            requestList,
+            requestHandler,
+            errorHandler,
+            failedRequestHandler,
+        });
+
+        await requestList.initialize();
+        await basicCrawler.run();
```
B4nan, line 304: random thought: i saw this pattern in tests quite often, maybe we could run `requestList.initialize()` automatically in `crawler.run()`, wdyt @mnmkng?

packages/basic-crawler/src/internals/basic-crawler.ts
```diff
@@ -150,6 +152,24 @@ export interface BasicCrawlerOptions<
...
+     * and would be retried by the crawler.
+     *
+     * The function receives the following object as the first argument:
+     * ```
+     * {
+     *     request: Request,
+     *     session: Session,
+     *     crawler: BasicCrawler,
+     * }
+     * ```
```
B4nan, line 166: same as the other comment, we can just link the interface (btw there can be more things right, e.g. proxy configuration/info)

in fact this should be all available already via the API docs thru the `ErrorHandler` type

B4nan (approved):
looking good, few small nits

packages/basic-crawler/src/internals/basic-crawler.ts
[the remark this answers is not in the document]
B4nan: its a bit weird those two functions have different signatures, we have the error object as part of the failed request context, and here it is in the second arg. 

i would be fine having it in the context and sharing the exact same type as current `failedRequestHandler`, those two should be really the same, the difference is only in when they are executed

packages/basic-crawler/src/internals/basic-crawler.ts
```diff
@@ -150,6 +150,26 @@ export interface BasicCrawlerOptions<
...
+     *
+     * The function receives the following object as an argument:
+     * ```
+     * {
+     *     request: Request,
+     *     error: Error,
+     *     session: Session,
+     *     crawler: BasicCrawler,
+     * }
+     * ```
```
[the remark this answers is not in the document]
B4nan: > I am not sure whether it's even possible to use generics in JSDoc

Why would you want to use generics in jsdoc, we are in TS, just use generics. (btw the anser is yes, you have the `@template` tag for that, but its very limited, and the use case is to type JS code, we are in TS already)

Sounds like X/Y problem, so tell me first what would you like to do :]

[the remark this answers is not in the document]
AndreyBykov: @B4nan Or are you ok to have it just a normal text - without a link to specific context? Something like `The function receives the CralingContext as the first argument...` and leave the rest to TS types?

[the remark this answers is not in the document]
AndreyBykov: Ok - got it - will update this 👌

packages/basic-crawler/src/internals/basic-crawler.ts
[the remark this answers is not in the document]
mnmkng: Oh, but then, wouldn't it make sense to separate the args even in the `failedRequestHandler`. I'm 100% for them having to be the same. But to me it's a bit weird to have to create an extended type for the error handlers. Historically, there was no consistent crawling context, so we've added stuff quite liberally. But now that there is, I wonder if it might be better to do it differently. But I'll leave that up to you.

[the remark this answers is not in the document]
B4nan: On the other hand, in a fail handler, you will definitely want to access the error - so you might need the second parameter, but not the first one (as you might not care about the context).

test/core/crawlers/basic_crawler.test.ts
mnmkng: Actually, `openRequestList` does not exist anymore, right? 😄

AndreyBykov: I actually put it there just because it's kinda widely used in the tests, so I thought maybe there's some deeper meaning into this inside of the tests :) Will update it (at least here).

B4nan: #157

AndreyBykov: @B4nan I was just cleaning up some tests (basically doing the same) =D Need to go for like an hour... Will finish with tests right after and it should be it :)

AndreyBykov:
@B4nan all green finally. If you could have a final look... And that should be it 👌

B4nan (approved):

PR #152, opened 2022-06-29
fix: unref worker to prevent process from hanging when it can exit

### LINKED ISSUE #147: `MemoryStorage` makes the process stuck
**Describe the bug**
When `MemoryStorage` is initialized (its constructor is called), it starts the worker and it never finishes, making the main process stuck.

**To Reproduce**
Just use anything with the memory storage backend, e.g. try to read the input.

**Expected behavior**
Process should finish, it gets stuck instead.

vladfrangu:
Closes #147


B4nan (approved):

mnmkng (commented):
Lol the process.exits 😄

AndreyBykov (approved):

PR #155, opened 2022-06-30
fix: `userAgent` overrides `useFingerprints`

### LINKED ISSUE #113: PuppeteerCrawler/PlaywrightCrawler + user-provided user-agent + fingerprints behaviour is not consistent
**Describe the bug**
`PuppeteerCrawler` used with fingerprints **disabled** and user-provided user-agent shows warning that automatic fingerprint injection will be disabled:
`INFO  PuppeteerCrawler: Disabling automatic fingerprint injection because custom user agent has been provided.`
Otherwise works as expected.

`PlaywrightCrawler` used with fingerprints **enabled** and user-provided user-agent shows ingores the user-provided user-agent and does not disable automatic fingerprint injection.

**To Reproduce**
```
const crawler = new PuppeteerCrawler({
// const crawler = new PlaywrightCrawler({
    async requestHandler({ page, log }) {
        page.on('request', request => console.log(request.headers()));
        await page.reload();
    },
    browserPoolOptions: { useFingerprints: false },
    // browserPoolOptions: { useFingerprints: true },
    launchContext: {
        userAgent: 'test-user-agent',
        launchOptions: { headless: true },
    }
});

await crawler.addRequests(['https://www.example.com']);
await crawler.run();
```

**Expected behavior**
In the first case, we should remove the warning.
In the second case, we should probably disable fingerprint injection and show the warning, and thus - use the user-provided user-agent.

**System information:**
 - OS: MacOS 12.4
 - Node.js version: 16.14.0
 - Apify SDK version: 3.0.0.alpha

barjin:
closes #113 

PlaywrightCrawler and PuppeteerCrawler behavior standardized. Specifying `launchContext.userAgent` overrides the `browserPool.useFingerprints` setting - the warning message only shows when `browserPool.useFingerprints` is set, though.

### DIFF
--- a/packages/browser-crawler/src/internals/browser-crawler.ts
+++ b/packages/browser-crawler/src/internals/browser-crawler.ts
@@ -424,9 +424,9 @@ export abstract class BrowserCrawler<
             this.persistCookiesPerSession = false;
         }
 
-        if (launchContext?.userAgent && browserPoolOptions.useFingerprints) {
+        if (launchContext?.userAgent) {
+            if (browserPoolOptions.useFingerprints) this.log.info('Custom user agent provided, disabling automatic browser fingerprint injection!');
             browserPoolOptions.useFingerprints = false;
-            this.log.info('Custom user agent provided, disabling automatic browser fingerprint injection!');
         }
 
         const { preLaunchHooks = [], postLaunchHooks = [], ...rest } = browserPoolOptions;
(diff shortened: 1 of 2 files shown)


B4nan:
e2e are looking good, but we need to adjust some assertions in unit tests apparently.

barjin:
TIL fingerprints are enabled by default, explicit `browserPool.useFingerprint: true` gets overridden, but the default doesn't

PR #159, opened 2022-06-30
chore: rename get/setPuppeteerCookies to get/setCookies

### LINKED ISSUE #89: Consider renaming `Session#getPuppeteerCookies` method; replace puppeteer cookie types with our definition that matches both puppeteer and playwright
maybe we should rename the `getPuppeteerCookies` method? also looking at the code, we depend on puppeteer types in core because of it, we will need to fix that too

(let's not deal with that in this PR, just a thought)

_Originally posted by @B4nan in https://github.com/apify/apify-ts/pull/83#discussion_r858791717_

vladfrangu:
Closes #89

### DIFF
--- a/docs/upgrading/upgrading_v3.md
+++ b/docs/upgrading/upgrading_v3.md
@@ -125,6 +125,10 @@ const crawler = new PlaywrightCrawler({
 });
  ```
 
+## Session cookie method renames
+
+Previously, if you wanted to get or add cookies for the session that would be used for the request, you had to call `session.getPuppeteerCookies()` or `session.setPuppeteerCookies()`. Since this method could be used for any of our crawlers, not just PuppeteerCrawler, the methods have been renamed to `session.getCookies()` and `session.setCookies()` respectively. Otherwise, their usage is exactly the same!
+
 ## Memory storage
 
 When we store some data or intermediate state (like the one `RequestQueue` holds), we now use `@crawlee/memory-storage` by default. It is an alternative to the `@apify/storage-local`, that stores the state inside memory (as opposed to SQLite database used by `@apify/storage-local`). While the state is stored in memory, it also dumps it to the file system so we can observe it, as well as respects the existing data stored in KeyValueStore (e.g. the `INPUT.json` file).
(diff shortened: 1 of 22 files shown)


packages/core/src/session_pool/session_utils.ts
```diff
@@ -1,13 +1,12 @@
...
 import type { IncomingMessage } from 'node:http';
-import type { HTTPResponse } from 'puppeteer';
 import { Cookie } from 'tough-cookie';
 import { CookieParseError } from './errors';
 
 /**
  * @internal
  */
-export function getCookiesFromResponse(response: IncomingMessage | HTTPResponse | { headers: Dictionary<string | string[]> }): Cookie[] {
+export function getCookiesFromResponse(response: IncomingMessage | BrowserLike | { headers: Dictionary<string | string[]> }): Cookie[] {
```
B4nan, line 9: maybe `BrowserLikeResponse` would be better name?

B4nan (approved):
lets add this to the list of breaking changes, both in upgrading guide and changelog.md

packages/core/src/session_pool/session_utils.ts
vladfrangu: Sure

vladfrangu:
Is this good? I think I documented it well enough but would love feedback 👀

docs/upgrading/upgrading_v3.md
```diff
@@ -125,6 +125,10 @@ const crawler = new PlaywrightCrawler({
 });
  ```
 
+## Session cookie method renames
+
+Previously, if you wanted to get or add cookies for the session that would be used for the request, you had to call `session.getPuppeteerCookies()` or `session.setPuppeteerCookies()`. Since this method could be used for any of our crawlers, not just PuppeteerCrawler, the methods have been renamed to `session.getCookies()` and `session.setCookies()` respectively. Otherwise, their usage is exactly the same!
```
B4nan, line 130:
```suggestion
Previously, if we wanted to get or add cookies for the session that would be used for the request, we had to call `session.getPuppeteerCookies()` or `session.setPuppeteerCookies()`. Since this method could be used for any of our crawlers, not just `PuppeteerCrawler`, the methods have been renamed to `session.getCookies()` and `session.setCookies()` respectively. Otherwise, their usage is exactly the same!
```

PR #165, opened 2022-07-04
feat: add setStatusMessage for setting custom run status

barjin:
Links to API docs are very general, as the endpoint is not documented yet

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4528,9 +4528,9 @@
             "link": true
         },
         "node_modules/apify-client": {
-            "version": "2.6.0-beta.0",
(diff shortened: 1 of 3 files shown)


B4nan (changes requested):
you will need to bump the client version and regenerate the lock flie

barjin:
just making sure - will it work when the required version is a [`-beta.0` prerelease](https://www.npmjs.com/package/apify-client/v/2.6.0-beta.0)? Do we want to bump the dependency version to beta?

B4nan:
yeah its fine for now, i will ship stable versions of everything before we ship stable of crawlee. we might want to merge some other things to the client so lets not release it to stable right now

PR #170, opened 2022-07-07
docs: last major portion of guide updates

AndreyBykov:
Mainly:
- env vars guide
- apify platform guide.

Would appreciate a feedback on both. For env vars - that I did not mess up anything in CRAWLEE_/APIFY_ part. For platform - I tried to make sure all previous platform related content is there (and not in the other guide). But still - order,  maybe I missed something etc.

There are several smaller things I need to take care for other guides (some examples, update gifs, you -> we), but I would probably do it in next PR and probably after API docs check as it's really non-important stuff I'd say and it's just a very few places.

Few random things I noticed while checking the examples etc:
1. `APIFY_` vs `CRAWLEE_` env vars -> if I set `process.env.CRAWLEE_LOG_LEVEL` locally to `DEBUG` - it would not show debug messages. With `log.setLevel(log.LEVELS.DEBUG);` it works. Is it expected to be this way?
2. If I am logged out of APIFY account locally and create a configuration with `const sdk = new Actor({ token: 'xxx' })` with proper token and then try to `sdk.createProxyConfiguration()` it would fail - am I doing something wrong or is it a bug? I would expect that creating an instance while explicitly providing the token should then work with proxy (and maybe other places where the token is used).

### DIFF
--- a/docs/guides/apify_platform.mdx
+++ b/docs/guides/apify_platform.mdx
@@ -130,12 +130,12 @@ Now (assuming we are already logged in to our Apify account) we can easily deplo
 apify push
 ```
 
-Our script will be uploaded to the Apify platform and built there so that it can be run. For more information, view the
+Our script will be uploaded to and built on the Apify platform so that it can be run there. For more information, view the
 [Apify Actor](https://docs.apify.com/cli) documentation.
 
 ## Usage on Apify platform
 
-We can also develop our actor in an online code editor directly on the platform (we'll need an Apify Account). Let's go to the [Actors](https://console.apify.com/actors) page in the app, click *Create new* and then go to the *Source* tab and start writing our code or paste one of the examples from the Examples section.
+We can also develop our actor in an online code editor directly on the platform (we'll need an Apify Account). Let's go to the [Actors](https://console.apify.com/actors) page in the app, click *Create new* and then go to the *Source* tab and start writing our code or paste one of the examples from the [Examples](../examples) section.
 
 ## Storages
 
@@ -145,10 +145,18 @@ There are several things worth mentioning here.
 2. In order to open the storage, we shouldn't use the storage classes, but instead use the Actor class. Thus, instead of <ApiLink to="core/class/KeyValueStore#open">`KeyValueStore.open()`</ApiLink>, <ApiLink to="core/class/Dataset#open">`Dataset.open()`</ApiLink> and <ApiLink to="core/class/RequestQueue#open">`RequestQueue.open()`</ApiLink>, we could use <ApiLink to="apify/class/Actor#openKeyValueStore">`Actor.openKeyValueStore()`</ApiLink>, <ApiLink to="apify/class/Actor#openDataset">`Actor.openDataset()`</ApiLink> and <ApiLink to="apify/class/Actor#openRequestQueue">`Actor.openRequestQueue()`</ApiLink> respectively. Using each of these methods allows us to pass the <ApiLink to="apify/interface/OpenStorageOptions">`OpenStorageOptions`</ApiLink> as a second argument, which has only one optional property: <ApiLink to="apify/interface/OpenStorageOptions#forceCloud">`forceCloud`</ApiLink>. If set to `true` - cloud storage will be used instead of the folder on the local disk.
 3. When the <ApiLink to="core/class/Dataset">`Dataset`</ApiLink> is stored on the `Apify platform`, we can export its data to the following formats: HTML, JSON, CSV, Excel, XML and RSS. The datasets are displayed on the actor run details page and in the [Storage](https://console.apify.com/storage) section in the Apify Console. The actual data is exported using the [Get dataset items](https://apify.com/docs/api/v2#/reference/datasets/item-collection/get-items) Apify API endpoint. This way you can easily share the crawling results.
 
+**Related links**
+
+- [Apify platform storage documentation](https://docs.apify.com/storage)
+- [View storage in Apify Console](https://console.apify.com/storage)
+- [Key-value stores API reference](https://apify.com/docs/api/v2#/reference/key-value-stores)
+- [Datasets API reference](https://docs.apify.com/api/v2#/reference/datasets)
+- [Request queues API reference](https://docs.apify.com/api/v2#/reference/request-queues)
+
 ## Important environment variables
 
-The following environment variables have large impact on the way Apify works and its behavior
-can be changed significantly by setting or unsetting them.
+The following environment variables have large impact on the way Actors work, and their behavior
+can be changed significantly by setting or unsetting these vars.
 
 ### `APIFY_TOKEN`
 
@@ -219,7 +227,7 @@ const proxyConfiguration = await Actor.createProxyConfiguration();
 const proxyUrl = await proxyConfiguration.newUrl();
 ```
 
-Note that unlike with Crawlee, we shouldn't use the constructor to create <ApiLink to="core/class/ProxyConfiguration">`ProxyConfiguration`</ApiLink> instance. For using Apify Proxy we should create an instance using the <ApiLink to="apify/class/Actor#createProxyConfiguration">`Actor.createProxyConfiguration()`</ApiLink> function instead.
+Note that unlike using proxy in Crawlee, we shouldn't use the constructor to create <ApiLink to="core/class/ProxyConfiguration">`ProxyConfiguration`</ApiLink> instance. For using Apify Proxy we should create an instance using the <ApiLink to="apify/class/Actor#createProxyConfiguration">`Actor.createProxyConfiguration()`</ApiLink> function instead.
 
 ### Apify Proxy vs. Own proxies
 
@@ -255,9 +263,4 @@ in the [proxy dashboard](https://console.apify.com/proxy).
 
 **Related links**
 
-- [Apify platform storage documentation](https://docs.apify.com/storage)
-- [View storage in Apify Console](https://console.apify.com/storage)
-- [Key-value stores API reference](https://apify.com/docs/api/v2#/reference/key-value-stores)
-- [Datasets API reference](https://docs.apify.com/api/v2#/reference/datasets)
-- [Request queues API reference](https://docs.apify.com/api/v2#/reference/request-queues)
 - [Apify Proxy docs](https://docs.apify.com/proxy)
--- a/docs/guides/apify_platform.mdx
+++ b/docs/guides/apify_platform.mdx
@@ -73,6 +73,13 @@ Actors can be shared in the [Apify Store](https://apify.com/store) so that other
 But don't worry, if you share your actor in the store and somebody uses it, it runs under their account,
 not yours.
 
+**Related links**
+
+- [Store of existing actors](https://apify.com/store)
+- [Documentation](https://docs.apify.com/actors)
+- [View actors in Apify Console](https://console.apify.com/actors)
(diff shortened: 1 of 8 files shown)


AndreyBykov:
Or I could actually quickly fix the last things I mentioned above during/after the review (as it's really just a very few places).

AndreyBykov:
Re CRAWLEE_LOG_LEVEL - created https://github.com/apify/apify-ts/issues/171 not to forget.

As for the token/configuration: did not try it with v2 (and I doubt many people were using it without logging in to apify) as it was kinda intuitive to install Apify and be logged in, but I guess it would fail too (will check in a sec). With v3:
```
import { Actor } from 'apify';

const sdk = new Actor({ token: 'xxxxxx' });
const proxyConfiguration = await sdk.createProxyConfiguration();
```

fails with `Error: Apify Proxy password must be provided using options.password or the "APIFY_PROXY_PASSWORD" environment variable. If you add the "APIFY_TOKEN" environment variable, the password will be automatically inferred`

Starting with `npm start`.

AndreyBykov:
Actually - with v2 it works as expected:

```
const { Apify } = require('apify');

const sdk = new Apify({ token: 'xxxxxx' });

sdk.main(async () => {
    const proxyConfiguration = await sdk.createProxyConfiguration({ groups: ['SHADER'] });
    console.log(proxyConfiguration.newUrl());
});
```

It correctly logs my proxy password and creates proxyUrl, while if I would not provide the token - it also fails. Also starting it with `npm start`.

I thought that the problem with v3 was that I don't do Actor.init(). Tried to add it, but it still fails.

docs/guides/apify_platform_init_exit.ts
```diff
@@ -0,0 +1,30 @@
...
+        // Save extracted data to dataset.
+        await Actor.pushData({ url, title });
+    },
+});
+
+// Choose the first URL to open.
+await crawler.addRequests([{ url: 'https://www.iana.org/' }]);
+
+// Start the crawler.
+await crawler.run();
```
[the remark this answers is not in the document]
AndreyBykov: So I guess I'll update it in some other examples. But I guess we should leave both options visible (in guides/examples)?

docs/guides/typescript_project.mdx
```diff
@@ -99,7 +99,7 @@ Caveats
 ========
 
 As of version 1.0+, the generated typings, due to JSDoc limitations, have some properties
-and parameters annotated with `any` type, therefore the settings `noImplicitAny` and `strictNullChecks`, set to `true`, may not be advised. You may try enabling them, but it might hinder development because of the need for typecasts to be able to compile, your mileage may vary.
+and parameters annotated with `any` type, therefore the settings `noImplicitAny` and `strictNullChecks`, set to `true`, may not be advised. They could be enabled, but it might hinder development because of the need for typecasts to be able to compile, and thus our mileage may vary.
```
B4nan, line 102: this whole TS guide is very outdated, will try to find some time for it after i am done with the getting started guide

AndreyBykov:
@B4nan So - besides the review/comments:
- need to update the links in `apify_platform` guide to the new repo (could do within this PR after actor SDK is moved there or separately later)
- `avoid_blocking` guide - I guess will update after https://github.com/apify/apify-ts/pull/169 is merged. Wanna add an example for puppeteer (which would basically be the same as playwright, so super fast to do) + would change the text and mention that fingerprints are enabled by default, how to disable + keep the details about `fingerprintGeneratorOptions`, but this part would also be fast.
- `quick_start` - wanted to update/add new gifs here (for cheerio/playwright/puppeteer). The question is - what's the top file size we want there? Currently we have a gif with ~900kb, 705x529 resolution there, ~10 sec duration. I tried to create one and it's either also a bit blurry, or becomes quite heavy. Should I take care of it or should I just keep the current gif and ditch playwright/cheerio?

Otherwise waiting for commments. Tomorrow will go through the API docs.

docs/guides/docker_images_ts
```diff
@@ -6,12 +6,13 @@ COPY . ./
 RUN npm install --include=dev \
     && npm run build
 
-# create final image, copy only package.json, readme and compiled code
+# create final image
 FROM apify/actor-node:16
 # copy only necessary files
-COPY --from=builder /usr/src/app/package*.json ./
+COPY --from=builder /usr/src/app/package.json ./
```
vladfrangu, line 12: This line should probably be kept as is so it also copies the package-lock file

docs/guides/typescript_project.mdx
```diff
@@ -113,14 +113,15 @@ interface MySchema {
     expectedParam2?: number;
 }
 
-const input: MySchema = (await KeyValueStore.getInput()) as any; // getInput returns Promise<Object<string, *>|string|Buffer|null> here
+// getInput returns Promise<Object<string, *>|string|Buffer|null> here
+const input: MySchema = (await KeyValueStore.getInput()) as any;
```
vladfrangu, line 117:
```suggestion
const input = await KeyValueStore.getInput<MySchema>();
```

docs/guides/apify_platform.mdx
```diff
@@ -5,75 +5,261 @@ title: Apify Platform
...
+
+- [Apify platform storage documentation](https://docs.apify.com/storage)
+- [View storage in Apify Console](https://console.apify.com/storage)
+- [Key-value stores API reference](https://apify.com/docs/api/v2#/reference/key-value-stores)
+- [Datasets API reference](https://docs.apify.com/api/v2#/reference/datasets)
+- [Request queues API reference](https://docs.apify.com/api/v2#/reference/request-queues)
+
+## Environment variables
+
+The following are some additional environment variables specific to Apify platform. More Crawlee specific environment variables could be found in the [Environment Variables](../guides/environment-variables) guide.
```
vladfrangu, line 168: Doesn't this also work?

```suggestion
The following are some additional environment variables specific to Apify platform. More Crawlee specific environment variables could be found in the [Environment Variables](./environment-variables) guide.
```

docs/guides/quick_start.mdx
```diff
@@ -13,14 +13,14 @@ import CheerioSource from '!!raw-loader!./quick_start_cheerio.ts';
...
-tutorial that will take you step by step through creating your first scraper.
+This short tutorial will set us up to start using Crawlee in a minute or two,
+while [Getting Started](../guides/getting-started) is a comprehensive step-by-step guide for creating the first scraper.
+So let's check out this page first and then proceed to the latter to learn more.
 
 ## Local stand-alone usage
 
 Crawlee requires [Node.js](https://nodejs.org/en/) 16 or later.
-Add Crawlee to any Node.js project by running:
+It could be added to any Node.js project by running:
```
vladfrangu, line 23:
```suggestion
It can be added to any Node.js project by running:
```

docs/guides/quick_start.mdx
```diff
@@ -90,22 +90,35 @@ Add PlaywrightCrawler gif here.
...
+- [Request storage](../guides/request-storage)
+- [Result storage](../guides/result-storage)
 
 ## Local usage with Crawlee command-line interface (CLI)
 
-To create a boilerplate of your project, you can use the [Crawlee command-line interface (CLI)](https://...) tool by running:
+To create a boilerplate of our project, we can use the [Crawlee CLI](https://www.npmjs.com/package/@crawlee/cli) tool by running:
+
+```bash
+npx crawlee create my-cheerio-crawler
```
vladfrangu, line 108: Does this actually point to the right tool? I know in the case of apify-cli I had to do `npx apify-cli` instead 😅

docs/upgrading/upgrading_v3.md
```diff
@@ -215,7 +215,7 @@ await result.waitForAllRequestsToBeAdded;
 
 ## Less verbose error logging
 
-Previously an error throw from inside request handler resulted in full error object being logger. With crawlee, we log only the error message as a warning as long as we know the request will be retried. If you want to enable verbose logging as in v2, use `CRAWLEE_VERBOSE_LOG` env var.
+Previously an error thrown from inside request handler resulted in full error object being logged. With crawlee, we log only the error message as a warning as long as we know the request will be retried. If you want to enable verbose logging as in v2, use `CRAWLEE_VERBOSE_LOG` env var.
```
vladfrangu, line 218: Is there a non-env variable way to achieve this too?

```suggestion
Previously an error thrown from inside request handler resulted in full error object being logged. With Crawlee, we log only the error message as a warning as long as we know the request will be retried. If you want to enable verbose logging like in v2, use the `CRAWLEE_VERBOSE_LOG` env var.
```

vladfrangu (changes requested):
Just some small nits

docs/guides/quick_start.mdx
B4nan: Yes it does, the metapackage defines a CLI script too.

https://github.com/apify/apify-ts/blob/master/packages/crawlee/package.json#L9

docs/upgrading/upgrading_v3.md
B4nan: Currently not, but we could have one.

docs/guides/docker_images.mdx
```diff
@@ -165,11 +150,28 @@ pre-installed.
...
 Same idea as [`actor-node-playwright-chrome`](#actor-node-playwright-chrome), but with WebKit
 pre-installed.
 
 ```dockerfile
 FROM apify/actor-node-playwright-webkit:16
 ```
+
+## Example Dockerfile
+
+To use the above images, it's necessary to have a [`Dockerfile`](https://docs.docker.com/engine/reference/builder/). We can either use this example, or bootstrap our projects with the [Apify CLI](./getting_started.mdx#creating-a-new-project) which automatically copies the correct Dockerfile into our project folder.
```
B4nan, line 164:
```suggestion
To use the above images, it's necessary to have a [`Dockerfile`](https://docs.docker.com/engine/reference/builder/). We can either use this example, or bootstrap our projects with the [Crawlee CLI](./getting_started.mdx#creating-a-new-project) which automatically adds the correct Dockerfile into our project folder.
```

docs/guides/docker_images_ts
```diff
@@ -6,12 +6,13 @@ COPY . ./
...
 COPY --from=builder /usr/src/app/README.md ./
 COPY --from=builder /usr/src/app/dist ./dist
+COPY --from=builder /usr/src/app/apify.json ./apify.json
```
B4nan, line 15:
```suggestion
```

B4nan:
Please address the feedback in another PR, I need to merge this now as the website is live and the PR will only make things better even if incomplete :]

packages/apify/src/configuration.ts
```diff
@@ -24,20 +24,27 @@ export interface ConfigurationOptions extends CoreConfigurationOptions {
...
+ *
  * 2. To get the global configuration (singleton instance). It will respect the environment variables.
- *   ```js
- *   import { BasicCrawler, Configuration } from 'crawlee';
  *
- *   const config = new Configuration({ persistStateIntervalMillis: 30_000 });
- *   const crawler = new BasicCrawler({ ... }, config);
- *   ```
+ *    ```javascript
+ *    import { BasicCrawler, Configuration } from 'crawlee';
```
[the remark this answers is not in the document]
AndreyBykov: @B4nan I actually did not change the examples here - I just updated the formatting as the spacing was not correct and thus it was not rendered properly + added a missing import.

But yeah - noted - will do together with other docs fixes.

PR #172, opened 2022-07-08
feat: prevent hybrid plugins from being used with browser-pool

### LINKED ISSUE #156: One driver per browser pool instance
As per https://github.com/apify/apify-js/pull/1228#discussion_r744304601

vladfrangu:
Closes #156 

Makes the browser-pool ensure that the constructor of the first plugin is whats used for all provided plugins (making it compatible with user-provided plugins too...should there be any)

### DIFF
--- a/docs/upgrading/upgrading_v3.md
+++ b/docs/upgrading/upgrading_v3.md
@@ -19,7 +19,7 @@ Moreover, the Crawlee library is published as several packages under `@crawlee` 
 - `@crawlee/core`: the base for all the crawler implementations, also contains things like `Request`, `RequestQueue`, `RequestList` or `Dataset` classes
 - `@crawlee/basic`: exports `BasicCrawler`
 - `@crawlee/cheerio`: exports `CheerioCrawler`
-- `@crawlee/browser`: exports `BrowserCrawler`
+- `@crawlee/browser`: exports `BrowserCrawler` (which is used for creating `@crawlee/playwright` and `@crawlee/puppeteer`)
 - `@crawlee/playwright`: exports `PlaywrightCrawler`
 - `@crawlee/puppeteer`: exports `PuppeteerCrawler`
 - `@crawlee/memory-storage`: `@apify/storage-local` alternative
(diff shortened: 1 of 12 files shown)


packages/browser-pool/src/browser-pool.ts
```diff
@@ -327,6 +327,24 @@ export class BrowserPool<
...
+
+            if (!(providedPlugin instanceof firstPluginConstructor)) {
+                const firstPluginName = firstPluginConstructor.name;
+                const providedPluginName = (providedPlugin as BrowserPlugin).constructor.name;
+
+                throw new Error(
+                    [
+                        `Browser plugin at index ${i} (${providedPluginName}) is not an instance of the same plugin`,
+                        `as the first plugin provided (${firstPluginName}).`,
+                    ].join(' '),
```
B4nan, line 343: this looks weird :] i'd just add eslint ignore comment here and keep it all on single line

B4nan (approved):
lets add this to the list of BCs

vladfrangu:
Where...exactly should I mention this BC? In the root CHANGELOG.md?

B4nan:
Yes, the changelog and [here](https://github.com/apify/apify-ts/blob/master/docs/upgrading/upgrading_v3.md#smallerinternal-breaking-changes), not sure if they are out of sync.

szmarczak (approved):
lgtm

packages/browser-pool/test/proxy-sugar.test.ts
```diff
@@ -0,0 +1,63 @@
+import { BrowserPool, PuppeteerPlugin, PlaywrightPlugin } from '@crawlee/browser-pool';
+import http from 'node:http';
+import type { AddressInfo } from 'node:net';
+import puppeteer from 'puppeteer';
+import playwright from 'playwright';
+import type { Server as ProxyChainServer } from 'proxy-chain';
+import { promisify } from 'node:util';
+import { createProxyServer } from '../../../test/browser-pool/browser-plugins/create-proxy-server';
```
szmarczak, line 8: Isn't the `../` moot?

vladfrangu: Probably, I forget we have some aliases setup

vladfrangu: Nope, doesn't work... I think its because theres no mapping in the folder its in

packages/browser-pool/src/index.ts
```diff
@@ -54,3 +54,9 @@ export type {
     GetFingerprintReturn,
 } from './fingerprinting/types';
 export type { InferBrowserPluginArray } from './utils';
+export type {
+    PuppeteerController,
+} from './puppeteer/puppeteer-controller';
+export type {
+    PlaywrightController,
+} from './playwright/playwright-controller';
```
B4nan, line 62: this will conflict with the classes in pw and pup crawler packages, or not?

vladfrangu: We don't have any class called Puppeteer/PlaywrightController in them (and the build tests pass) 🤷‍♂️

vladfrangu: Oh thooose. Yeah good catch, can delete those now 🎉

docs/upgrading/upgrading_v3.md
```diff
@@ -272,17 +272,27 @@ Last but not least, `retry.limit` is set to `0`. This is because `crawlee` has i
 
 For more info please refer to the Got [documentation](https://github.com/sindresorhus/got#documentation).
 
-### Removed options
+## Removed options
```
B4nan, line 275: i believe those should stay as they were. the whole block is about removal of requestAsBrowser, which is the second level heading

vladfrangu: This is how it looks now..., it was super broken before
Google Chrome - 2022-07-08 at 15 05 39@2x

vladfrangu: OH I see what you meant. Woooops. Will fix

B4nan:
green ci finally! :D feel free to merge

PR #173, opened 2022-07-08
docs: update LICENSE.md with commercial license terms

mnmkng:
In line with what we discussed, I'm updating the license to make use of Crawlee for commercial purposes in large companies subject to a separate license agreement.


B4nan (approved):

B4nan:
We should also adjust the readme

https://github.com/apify/apify-ts/blob/master/README.md#license

PR #176, opened 2022-07-09
docs: few more updates to guides

AndreyBykov:
- fix apify console links in `apify-platform`
- fix links to docker guide in `docker-images`: instead of sdk.apify.com -> use crawlee.dev
- update `avoid-blocking` guide
- `quick-start` - instead of the gif for cheerio - added static log example + kept the same gif for playwright/puppeteer + added a sentence before gif/log. I think it should be ok but let me know if you think it's not.

Last thing to do on my list here is to update the `apify-platform` guide with links to new repo, so will do in another PR once `apify-sdk-js` is live.


B4nan:
Will merge to see how it looks rendered, so lets have another PR for feedback if any.

PR #179, opened 2022-07-12
docs: update guides + clean-up basic crawler api docs

AndreyBykov:
- Went throught `got-scpraing` guide and fixed some typos, `you -> we`, etc. Also move it after `avoid blocking` in the side menu.
- Went through Basic Crawler API docs, fixed/updated some descriptions, added some links, etc.

Please check it out if you would have the internet (maybe even render the docs) - if it's all good - will do the rest tomorrow (or better to say today :/)

### DIFF
--- a/docs/guides/got_scraping.mdx
+++ b/docs/guides/got_scraping.mdx
@@ -5,7 +5,7 @@ title: Got Scraping
 
 ## Intro
 
-If we're using Basic Crawler, we have to send requests manually. In order to do this, We need to call `sendRequest()` like this:
+When using `BasicCrawler`, we have to send the requests manually. In order to do this, we can use the context-aware `sendRequest()` function:
 
 ```ts
 import { BasicCrawler } from 'crawlee';
--- a/docs/guides/got_scraping.mdx
+++ b/docs/guides/got_scraping.mdx
@@ -150,7 +150,7 @@ const crawler = new BasicCrawler({
 
 This option specifies the maximum number of `Got` retries.
 
-By default, `retry.limit` is set to `0`. This is because `Crawlee` has its own (complicated enough) retry management.
+By default, `retry.limit` is set to `0`. This is because Crawlee has its own (complicated enough) retry management.
 
 We suggest NOT changing this value for stability reasons.
 
--- a/docs/guides/got_scraping.mdx
+++ b/docs/guides/got_scraping.mdx
@@ -1,13 +1,15 @@
 ---
 id: got-scraping
-title: Got Scraping guide
+title: Got Scraping
 ---
 
 ## Intro
 
-If you're using Basic Crawler, you have to send requests manually. In order to do this, you need to call `sendRequest()` like this:
+If we're using Basic Crawler, we have to send requests manually. In order to do this, We need to call `sendRequest()` like this:
 
 ```ts
+import { BasicCrawler } from 'crawlee';
+
 const crawler = new BasicCrawler({
     async requestHandler({ sendRequest, log }) {
         const res = await sendRequest();
@@ -17,7 +19,7 @@ const crawler = new BasicCrawler({
 ```
 
 It uses [`got-scraping`](https://github.com/apify/got-scraping) under the hood.
-Got Scraping is a [Got](https://github.com/sindresorhus/got) extension developed to mimic browser requests, so there's a high chance we'll open the webpage.
+Got Scraping is a [Got](https://github.com/sindresorhus/got) extension developed to mimic browser requests, so there's a high chance we'll open the webpage without getting blocked.
 
 ## `sendRequest` API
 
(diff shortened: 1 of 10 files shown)


docs/guides/got_scraping.mdx
```diff
@@ -1,13 +1,15 @@
 ---
 id: got-scraping
-title: Got Scraping guide
+title: Got Scraping
 ---
 
 ## Intro
 
-If you're using Basic Crawler, you have to send requests manually. In order to do this, you need to call `sendRequest()` like this:
+If we're using Basic Crawler, we have to send requests manually. In order to do this, We need to call `sendRequest()` like this:
```
B4nan, line 8:
```suggestion
When using `BasicCrawler`, we have to send the requests manually. In order to do this, we can use the context-aware `sendRequest()` function:
```

docs/guides/got_scraping.mdx
```diff
@@ -49,35 +51,37 @@ async sendRequest(overrideOptions?: GotOptionsInit) => {
 
 By default, it's the URL of current task. However you can override this with a `string` or a `URL` instance if necessary.
 
-See the [Got documentation](https://github.com/sindresorhus/got/blob/main/documentation/2-options.md#url) for more details.
+*More details: [Got documentation](https://github.com/sindresorhus/got/blob/main/documentation/2-options.md#url).*
```
B4nan, line 54: maybe its better this way? idk

```suggestion
*More details in [Got documentation](https://github.com/sindresorhus/got/blob/main/documentation/2-options.md#url).*
```

docs/guides/got_scraping.mdx
```diff
@@ -135,26 +142,29 @@ const crawler = new BasicCrawler({
...
+- *[Got documentation](https://github.com/sindresorhus/got/blob/main/documentation/2-options.md#cookiejar)*
+- *[Tough Cookie documentation](https://github.com/salesforce/tough-cookie#cookiejarstore-options)*
 
 ### `retry.limit`
 
-This option specifes the maximum number of Got retries.
+This option specifies the maximum number of `Got` retries.
 
-By default, `retry.limit` is set to `0`. This is because `crawlee` has its own (complicated enough) retry management.
+By default, `retry.limit` is set to `0`. This is because `Crawlee` has its own (complicated enough) retry management.
```
B4nan, line 153:
```suggestion
By default, `retry.limit` is set to `0`. This is because Crawlee has its own (complicated enough) retry management.
```

docs/guides/request_storage.mdx
```diff
@@ -145,4 +145,4 @@ import { purgeDefaultStorages } from 'crawlee';
 await purgeDefaultStorages();
 ```
 
-Calling this function will clean up the default request storage directory (and also the request list stored in default key-value store). This is a shortcut for running (optional) `purge` method on the <ApiLink to="types/interface/StorageClient">`StorageClient`</ApiLink> interface, in other words it will call the `purge` method of the underlying storage implementation we are currently using. In addition, this method will make sure the storage is purged only once for a given execution context, so it is safe to call it multiple times.
+Calling this function will clean up the default request storage directory (and also the request list stored in default key-value store). This is a shortcut for running (optional) `purge` method on the <ApiLink to="core/interface/StorageClient">`StorageClient`</ApiLink> interface, in other words it will call the `purge` method of the underlying storage implementation we are currently using. In addition, this method will make sure the storage is purged only once for a given execution context, so it is safe to call it multiple times.
```
B4nan, line 148: just curious, the link was not working before? because it was valid, i just added reexport yesterday to the core package. i guess it again (wrongly) marked the one in types packages as reexport? we should probably report this to typedoc, i'd say there will be some issue about this already

B4nan (approved):
looking good, i'll merge it right ahead and we can again improve later via another PR

docs/guides/got_scraping.mdx
AndreyBykov: I was thinking about this one, both sounds good to me, will update to `in` with next PR :)

docs/guides/request_storage.mdx
AndreyBykov: Yeah, with `types/interface/StorageClient` it returned `Page does not exist` or something like that :/

And yeah - in `types` it has it only in `References` as `Re-exports [StorageClient](https://crawlee.dev/api/core/interface/StorageClient)`

docs/guides/got_scraping.mdx
szmarczak, line 54: What's the rationale behind this change? I don't mind, just asking.

AndreyBykov: Originally @B4nan mentioned this `you` -> `we` thing on Slack, so basically I tried to change `[you] see more details` to something less direct/addressed 🤷‍♂️

But in the end, we're not doing it...

szmarczak: 😂 Ok I get it now :D That's kinda dumb because there's often `Read more` in other projects' docs.

Like I suggested, `you` refers explicitly to the developer reading the docs, so if *they* are given an option then the steps depend on *their* path *they* have chosen. So `you` would be appropriate. https://docs.oracle.com/en/java/

> Java SE lets you develop

There's an implied choice. You can choose Java or any other language.

OTOH `we` refers to both the developer and the project, there may be a thing that applies to both persons. For example a tutorial. https://docs.oracle.com/search/?q=we&category=java&product=en%2Fjava

> We will run the HelloWorld sample

AndreyBykov: I think that's exactly what we're doing in the end 😄 But this PR was already merged before :/

PR #180, opened 2022-07-13
docs: more updates to api docs - browser-crawler package

AndreyBykov:
There's only few tiny changes to `basic crawler` + `browser crawler` package should be done. I'm slow as hell today - probably should not go to sleep at 7 am 😄. I would rather push more changes (not to create 10 PRs for this). But it COULD be merged already, this `browser-crawler` package is done from my perspective.

One thing I wanted to ask - is it intentional that `browser-pool` is not exposed? I have replaced the `BrowserPool` links to the current `apify-ts` repo, but it would change. So I should either update the links later, or if `BrowserPool` will be shown in the api docs - I would replace it with `{@link BrowserPool}`.

One more thing not to forget - search on the site still does not work - is it expected?


B4nan:
Search does not work because algolia guys are slow as hell. I'm in contact with them.

B4nan:
Good catch about the browser pool, [added](https://github.com/apify/apify-ts/commit/d9df30ca65daef6b3d7bad45090ffd12f5f9add3) now to API docs.

packages/browser-crawler/src/internals/browser-crawler.ts
```diff
@@ -217,13 +222,14 @@ export interface BrowserCrawlerOptions<
     /**
      * Async functions that are sequentially evaluated after the navigation. Good for checking if the navigation was successful.
      * The function accepts `crawlingContext` as the only parameter.
-     * Example:
-     * ```
+     *
+     * @example
```
B4nan, line 226: not sure if we want to use the `@example` tag. docusaurus renders it [this way](https://crawlee.dev/api/basic-crawler/interface/CrawlerAddRequestsResult#waitForAllRequestsToBeAdded), might be better to keep it like before

also we use it exactly ones in the whole code base

AndreyBykov: Yeah, I saw it used in some parts, thougth that we decided to go this way at some point =D I do not like it either, so would revert. I do not like the `@default` either, but default values are usually something that does not take half of the screen + it makes it kinda more visible, so would probably leave defaults this way

B4nan: I hope the `@default` tag will get some love in the docusaurus plugin and will get rendered better in future versions.

AndreyBykov:
updated the links, the `example` tag, updated Actor links in apify-guide. Wanna merge it or I could continue here? @B4nan

B4nan:
do another PR, i want to have the docs as up to date as possible as they are live already

